PR #2, opened 2015-05-27
Add the TLD updater script for the Hostname validator

weierophinney:
ZF2 had a script, `bin/update_hostname_validator.php`, that could be run to update the list of valid TLDs in the `Hostname` validator. This patch ports it to `zend-validator`, where it now belongs. It required:

- Adding `zend-http` as a development dependency (as it uses `zend-http` to fetch the list of current valid TLDs from IANA).


Maks3w:
Can we automatize this?

At least to receive some alert when list is unsynced.

Finally. Update the list should be enough for to release a stable version inmediately

bin/update_hostname_validator.php
```diff
@@ -0,0 +1,130 @@
+<?php
```
Maks3w: Should this file to have a shebang and chmod +x?

It's what I expect for any bin file

Maks3w:
Could you add bin folder to .php_cs too?

bin/update_hostname_validator.php
weierophinney: I invoke it with PHP: `php bin/update_hostname_validator.php`; this has been how it's always been invoked since introduced to ZF2, and is only meant for updating the TLD list.

weierophinney:
> Can we automatize this?
> 
> At least to receive some alert when list is unsynced.

I'm really unsure how we can do that. If you have ideas, let me know, but I think any such effort is out-of-scope for the goals of this PR, which are simply to port the functionality into this repo from the main ZF2 repo.

Maks3w:
Probably the hostname list is an asset and should be managed from a different repo.

weierophinney:
> Could you add bin folder to .php_cs too?

Done.

Maks3w:
I agree my suggestions is not a "no go". Just try to improve now this part of code is under my eyes :)

weierophinney:
> Probably the hostname list is an asset and should be managed from a different repo.

We fetch the list from IANA; it's updated periodically, but they do not provide a feed to my knowledge.

weierophinney:
> Probably the hostname list is an asset and should be managed from a different repo.

We could potentially make this config-driven (i.e., inject the list when instantiating the validator), but that makes usage incredibly convoluted for many, and is more likely to introduce points of failure. Having it inlined in the class is a good solution, and we can update as we prepare to tag any new version.

Maks3w:
Automatization: Probably we can execute the binary after the CS command and validate against `git status` output

weierophinney:
> Automatization: Probably we can execute the binary after the CS command and validate against git status output

And do what? Commit the changes from Travis? That seems risky...

PR #4, opened 2015-06-04
bump minimum php version to 5.5

samsonasik:


gianarb:
This PR sounds good.. Why is it frozen?

samsonasik:
the 2nd commit can be cherry-picked for DateTimeInterface check

gianarb:
:+1:

Maks3w:
You can cherry pick the commit and create a PR only with that change.

samsonasik:
@Maks3w yes I can, but please, I create the PR before the dramatically change php require version for composer.json on all repo. I am wondering why all the change is not seeing the avaialble PR first :)

Maks3w:
@samsonasik it's simple. you did a batch of prs without coordinate with the crteam

gianarb:
All this words for a cherry-pick :D #25
Thanks @samsonasik for this contribution I don't know if this is a batch process but you have resolved all php5.5 compatibility and not only composer.json configuration.. Perfect!

baaahh :disappointed:

samsonasik:
we already did get info that zf 2.5 will require php >=5.5 ... so sorry for not asking about it first to crteam.

samsonasik:
and I check one by one the changes in every repo..., that ALL ok

PR #7, opened 2015-06-10
Add support for @param array Validator\File\Count\isValid

gustavopinho:


Maks3w:
Please add unit tests

src/File/Count.php
```diff
@@ -202,17 +202,22 @@ public function addFile($file)
...
-        }
-
-        if (($file !== null) && array_key_exists('tmp_name', $file)) {
-            $value = $file['destination'] . DIRECTORY_SEPARATOR . $file['name'];
-        }
-
-        if (($file === null) || !empty($file['tmp_name'])) {
-            $this->addFile($value);
+        if (is_string($value) && is_array($file)) {
+            $filename = $file['name'];
```
Maks3w: When is used $filename?

gustavopinho: When I add inputFilter
$fileInput->getValidatorChain()->attachByName('filecount', array('min' => 0, 'max' => 1));

Validator\Count\isValid @param $value receive 

array(5) {
  ["name"] => string(8) "file.png"
  ["type"] => string(9) "image/png"
  ["tmp_name"] => string(47) "\Temp\php436D.tmp"
  ["error"] => int(0)
  ["size"] => int(927150)
}

this cause

Warning: dirname() expects parameter 1 to be string, array given in ...Validator\File\Count.php on line 226

Then the validation does not work

PR #8, opened 2015-06-10
Allow long email addresses per rfc 5322

denis-sokolov:
The addresses are not only theoretical,
I have recently received emails from Trello and Peerby that exceed the 64 octet limit for the local part already.

### DIFF
--- a/src/EmailAddress.php
+++ b/src/EmailAddress.php
@@ -69,6 +69,7 @@ class EmailAddress extends AbstractValidator
         'useDeepMxCheck'    => false,
         'useDomainCheck'    => true,
         'allow'             => Hostname::ALLOW_DNS,
+        'strict'            => true,
         'hostnameValidator' => null,
     ];
 
@@ -78,6 +79,7 @@ class EmailAddress extends AbstractValidator
      * The following additional option keys are supported:
      * 'hostnameValidator' => A hostname validator, see Zend\Validator\Hostname
      * 'allow'             => Options for the hostname validator, see Zend\Validator\Hostname::ALLOW_*
+     * 'strict'            => Whether to adhere to strictest requirements in the spec
      * 'useMxCheck'        => If MX check should be enabled, boolean
      * 'useDeepMxCheck'    => If a deep MX check should be done, boolean
      *
@@ -499,7 +501,7 @@ class EmailAddress extends AbstractValidator
             return false;
         }
 
-        if ((strlen($this->localPart) > 64) || (strlen($this->hostname) > 255)) {
+        if ($this->getOption('strict') && (strlen($this->localPart) > 64) || (strlen($this->hostname) > 255)) {
             $length = false;
             $this->error(self::LENGTH_EXCEEDED);
         }
@@ -542,7 +544,13 @@ class EmailAddress extends AbstractValidator
     protected function idnToUtf8($email)
     {
         if (extension_loaded('intl')) {
-            return idn_to_utf8($email);
+            // The documentation does not clarify what kind of failure
+            // can happen in idn_to_utf8. One can assume if the source
+            // is not IDN encoded, it would fail, but it usually returns
+            // the source string in those cases.
+            // But not when the source string is long enough.
+            // Thus we default to source string ourselves.
+            return idn_to_utf8($email) ?: $email;
         }
         return $email;
     }
--- a/test/EmailAddressTest.php
+++ b/test/EmailAddressTest.php
@@ -293,6 +293,29 @@ class EmailAddressTest extends \PHPUnit_Framework_TestCase
         }
     }
 
+    /**
+     * Ensures that the validator follows expected behavior for valid email addresses with the non-strict option
+     *
+     * @return void
+     */
+    public function testNonStrict()
+    {
+        $validator = new EmailAddress(['strict' => false]);
+        $emailAddresses = [
+            // RFC 5321 does mention a limit of 64 for the username,
+            // but it also states "To the maximum extent possible,
+            // implementation techniques that impose no limits on the
+            // length of these objects should be used.".
+            // http://tools.ietf.org/html/rfc5321#section-4.5.3.1
+            'line length 320' => str_repeat('x', 309).'@domain.com',
+            'line length 321' => str_repeat('x', 310).'@domain.com',
+            'line length 911' => str_repeat('x', 900).'@domain.com',
+        ];
+        foreach ($emailAddresses as $input) {
+            $this->assertTrue($validator->isValid($input));
+        }
+    }
+
    /**
      * Ensures that the validator follows expected behavior for checking MX records
      *


src/EmailAddress.php
```diff
@@ -491,15 +491,17 @@ public function isValid($value)
...
+        if (!$this->splitEmailParts($value)) {
             $this->error(self::INVALID_FORMAT);
             return false;
         }
 
-        if ((strlen($this->localPart) > 64) || (strlen($this->hostname) > 255)) {
+        $this->hostname = $this->idnToUtf8($this->hostname);
+
+        // https://tools.ietf.org/html/rfc5322#section-2.1.1
+        if ((strlen($this->localPart) + strlen($this->hostname)) > 994) {
```
Maks3w: I guess `@` sign should be included to the length too.

denis-sokolov: I actually subtracted 4 from 998 to account for `To:` and `@`, but did not document it properly.
Thanks for noticing.

Maks3w:
@denis-sokolov The length of 998 is not applicable here. It's just the maximum line length in the SMTP protocol and does is not applicable to email address. (In fact there is no any length limit in the content, just fold the header in multiple lines and content is allowed to exceed 998 characters)

Please choose a better RFC as reference, if you can't find a reference just remove the local part constraint.

test/EmailAddressTest.php
```diff
@@ -286,7 +286,13 @@ public function testComplexLocalValid()
...
             '#Bob.Jones@domain.com',
             'Bob.Jones?@domain.com',
-            'Bob~Jones@domain.com'
+            'Bob~Jones@domain.com',
+            // RFC 5321 does mention a limit of 64 for the username,
+            // but it also states "To the maximum extent possible,
+            // implementation techniques that impose no limits on the
+            // length of these objects should be used.".
+            // http://tools.ietf.org/html/rfc5321#section-4.5.3.1
+            str_repeat('BobJones', 120).'@domain.com'
```
Maks3w: Instead use `'line length NN'  => str_repeat('x', NN) . '@x',`

Maks3w: NN will be more clear about what is the expected length

test/EmailAddressTest.php
```diff
@@ -451,7 +457,9 @@ public function testHostnameValidatorMessagesShouldBeTranslated()
     public function testEmailsExceedingLength()
     {
         $emailAddresses = [
-            'thislocalpathoftheemailadressislongerthantheallowedsizeof64characters@domain.com',
+            // RFC 5322 mentions a limit of 998 per header line.
+            // https://tools.ietf.org/html/rfc5322#section-2.1.1
+            str_repeat('bob', 330) . '@i.com',
```
Maks3w: same observation as above about how to write long cases.

test/EmailAddressTest.php
denis-sokolov: Thanks, updating.

denis-sokolov:
Thanks for the correction, @Maks3w.

denis-sokolov:
Finally, a passing test! Sorry for the noise - it passed locally initially, because I had no idn extension.

Please disregard the coveralls check, it's dumb. I removed 4 covered lines and the overall coverage now "fell". Stupid algorithm is stupid.

Maks3w:
@denis-sokolov Please read my comment above. Your length limit assumption is wrong. RFC 5321 is not a reference for this.

denis-sokolov:
@Maks3w, I did read it, it was helpful, thank you.
I think you were referring to the 998 limit of RFC 5322 that I did remove from the PR.

The remaining PR is about RFC 5321 (Internet Message Format) stating "To the maximum extent possible, implementation techniques that impose no limits on the limits of these objects should be used.".

src/EmailAddress.php
```diff
@@ -32,8 +31,7 @@ class EmailAddress extends AbstractValidator
         self::INVALID_SEGMENT    => "'%hostname%' is not in a routable network segment. The email address should not be resolved from public network",
         self::DOT_ATOM           => "'%localPart%' can not be matched against dot-atom format",
         self::QUOTED_STRING      => "'%localPart%' can not be matched against quoted-string format",
-        self::INVALID_LOCAL_PART => "'%localPart%' is not a valid local part for the email address",
-        self::LENGTH_EXCEEDED    => "The input exceeds the allowed length",
+        self::INVALID_LOCAL_PART => "'%localPart%' is not a valid local part for the email address"
```
Maks3w: Preserve the trailing comma

test/EmailAddressTest.php
```diff
@@ -286,7 +286,13 @@ public function testComplexLocalValid()
...
             '#Bob.Jones@domain.com',
             'Bob.Jones?@domain.com',
-            'Bob~Jones@domain.com'
+            'Bob~Jones@domain.com',
+            // RFC 5321 does mention a limit of 64 for the username,
+            // but it also states "To the maximum extent possible,
+            // implementation techniques that impose no limits on the
+            // length of these objects should be used.".
+            // http://tools.ietf.org/html/rfc5321#section-4.5.3.1
+            'line length 911' => str_repeat('x', 900).'@domain.com'
```
Maks3w: Do this test with 320 and 321 lengths instead. Reflect better the previous length barrier

Maks3w:
Yes. I was talking about the 998 line length limit

Maks3w:
At this moment I have a couple of concerns:

1. Seems we have two options of perform a validation. A relaxed one with length limits removal and strict one with the recommended length limits.

2.  The IDN to UTF-8 failure is enough for mark the address as invalid. Otherwise please open a bug ticket on PHP bug tracker.

3.

denis-sokolov:
> Seems we have two options of perform a validation. A relaxed one with length limits removal and strict one with the recommended length limits.
The distinction between strict and lax validation is important and is unfortunate.
I am using Zend mail module to parse emails, and that's the use case when I need lax validation.
The documentation [seems to suggest](http://framework.zend.com/manual/1.12/en/zend.mail.read.html) it's a [normal use case](http://framework.zend.com/manual/current/en/modules/zend.mime.message.html#parsing-a-string-to-create-a-zend-mime-message-object).
If I was to send emails, I would expect the validation to be strict.

> The IDN to UTF-8 failure is enough for mark the address as invalid. Otherwise please open a bug ticket on PHP bug tracker.
The same class a few lines above already [uses the fallback](https://github.com/zendframework/zend-validator/blob/master/src/EmailAddress.php#L532):
```php
return (idn_to_ascii($email) ?: $email);
```
If anything, the new change ensures consistency.

Maks3w:
I was suspecting this come from Zend\Mail :)

I think this is almost done.

Please add a option named `strict => true` for toggle on/off length validation.

Then we can create a PR against Zend\Mail turning the option to false.

denis-sokolov:
Done, @Maks3w.

PR #14, opened 2015-07-02
Remove min & max requirement from Between constructor

alexdenvir:
* Allows to be instantiated using default min and max values

### DIFF
--- a/src/Between.php
+++ b/src/Between.php
@@ -64,7 +64,13 @@ class Between extends AbstractValidator
         }
         if (!is_array($options)) {
             $options = func_get_args();
-            $temp['min'] = array_shift($options);
+
+            $temp = [];
+
+            if (!empty($options)) {
+                $temp['min'] = array_shift($options);
+            }
+
             if (!empty($options)) {
                 $temp['max'] = array_shift($options);
             }
--- a/src/Between.php
+++ b/src/Between.php
@@ -56,8 +56,6 @@ class Between extends AbstractValidator
      *   'inclusive' => boolean, inclusive border values
      *
      * @param  array|Traversable $options
-     *
-     * @throws Exception\InvalidArgumentException
      */
     public function __construct($options = null)
     {
--- a/src/Between.php
+++ b/src/Between.php
@@ -78,12 +78,6 @@ class Between extends AbstractValidator
             $options = $temp;
         }
 
-        if (count($options) !== 2
-            && (!array_key_exists('min', $options) || !array_key_exists('max', $options))
-        ) {
-            throw new Exception\InvalidArgumentException("Missing option. 'min' and 'max' have to be given");
-        }
-
         parent::__construct($options);
     }
 
--- a/test/BetweenTest.php
+++ b/test/BetweenTest.php
@@ -103,22 +103,6 @@ class BetweenTest extends \PHPUnit_Framework_TestCase
         $this->assertAttributeEquals($validator->getOption('messageVariables'), 'messageVariables', $validator);
     }
 
-    /**
-     * @covers Zend\Validator\Between::__construct()
-     * @dataProvider constructBetweenValidatorInvalidDataProvider
-     *
-     * @param array $args
-     */
-    public function testMissingMinOrMax(array $args)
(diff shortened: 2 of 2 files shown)


gianarb:
:1

alexdenvir:
This probably warrants a bit of an explanation, so here goes:

This validator currently throws an exception is you attempt to instantiate it without any options (in particular, we require a minimum of `min` and `max`). However, the validator already has default values for these options, which are not completely unreasonable.

It seems pretty logical to me to be able to call `new Between(['max' => 5]);` and not  have it throw an exception. I would also expect the service locator to be able to instantiate this validator if I wanted it to with the default options (ignoring the abuse of the service locator pattern).

I don't think this is a BC break, as anyone who is currently passing the min and max options through the constructor will still have it work

freax:
I think this PR needs some tests for sure `value` is in range in (0, PHP_INT_MAX) by default for future BC breaks.

alexdenvir:
Should point out that the test I've added is purely to test the options when different parameters are passed to the constructor. I'm haven't tested the value and output of isValid with the default options

manuakasam:
I don't want to put down any efforts to make the framework better and I appreciate this attempt as such.

However I am 100% against this PR.
I'm not even sure why there are default values for this Validator as they make no sense in likely 99.9% of all use-cases. The between validator - in MY personal opinion - should always be instantiated with the min/max values that you need. Relying on defaults here is just very prone to errors in my opinion and makes the validation-code much harder to understand, as you have to go inside the class to see the actual min/max values.

:-1: for me

alexdenvir:
@manuakasam I can see where you are coming from, however it is inconsistent with many of the other validators available in the package. As an example, take the InArray validator. You are able to create an instance without passing any options in to the constructor, but it is completely useless without setting the haystack as a minimum. That said, several other validators will also throw an exception if instantiated without appropriate options, so I guess you could consider this PR as testing the waters with regards to standardising :stuck_out_tongue: 

Ultimately the reason for me creating this PR is that I'm not a fan of being forced to pass an array of options in a constructor, as the values attached to array keys cannot be typehinted, there is no IDE autocomplete for the array keys themselves etc.

Compare calling:
```php
$validator = new Between(['min' => 5, 'max' => 10]);
```
to:
```php
$validator = new Between();
$validator->setMin(5)->setMax(10);
```
Any good IDE will autocomplete the function calls, while also telling you exactly what the function is doing and what value it is expecting, whether the value is typehinted etc.

gianarb:
I agree with  @manuakasam.. 
We can merge this PR for ZF2 but in the future for ZF3 we can remove default value..

alexdenvir:
Maybe the best solution going forward is to look at the way the Callback validator is used.

There is no default, the validator can be instantiated without passing in options into the constructor, but if you call isValid without a callback being set at all it throws an exception then. Any thoughts on that approach?

manuakasam:
To explain further. What I'd be ok with is to not have them pass within the
constructor but then remove the default values as such and return an
appropriate error message when there's no min max set.

My main issue is having default values for such a validator.

Alex Denvir <notifications@github.com> schrieb am Fr., 3. Juli 2015 um
13:02 Uhr:

Mezzle:
My opinion is that defaults shouldn't be there, we should be forced to set them, but I think throwing an exception here would be a BC break?

Might be worth it for future.  At the moment, @alexdenvir seems the best solution for what is there, as most validators don't force you to construct them in a specific way, whereas this does.

manuakasam:
I wasn't talking about an exception. it's simply that the isValid() should
return proper errors when both min and max are missing. Not sure if it's
already there or not.

Martin Meredith <notifications@github.com> schrieb am Fr., 3. Juli 2015 um
14:31 Uhr:

freax:
In my second thoughts may be it will be better to set `min` and `max` default values to "0" because of PHP_INT_MAX is a platform dependent number.

alexdenvir:
Closing this for now, as I've submitted an alternative pull request https://github.com/zendframework/zend-validator/pull/15 which should be considered instead of this

Issue #1, opened 2015-05-18
Road to ZF3


bakura10:
Hi,

I've done a massive PR to this (https://github.com/zendframework/zf2/pull/5067) but it's a massive work and a big hell to review.

I suggest that we:

1. Move validators to their specifc components (for instance moving Db validators to the Zend\Db component), and deprecating the validators here (for version 2.5).

2. Remove them in v3, and start the refactoring work. Here are various things I have done in the past:

* Make validators stateless (each validator returns a ValidationResult, that can be translated using another component for instance).
* Using the new service manager of ZF3.
* Homogeneize names (in ZF2, we didn't have time to use the "underscore_separated" convention for option name, which make it very incoherent with how the framework work).
* Various code cleaning, optimization for PHP 5.5

snapshotpl:
:+1: for stateless

Maks3w:
I suggest to split the proposal in invidual issues for to have clean discussion threads.

:+1: For place validator inside of each component.

weierophinney:
> Move validators to their specifc components (for instance moving Db validators to the Zend\Db component), and deprecating the validators here (for version 2.5).

First, v2.5 is imminent; I just need to ensure all the components are passing tests on current master, and then I'll be tagging their 2.5 versions, followed by the refactoring of the ZF2 repo to be a metapackage (see zendframework/zf2#7542 and [my own tree that extends that](https://github.com/weierophinney/zf2/tree/feature/7542)). As such, deprecation would have to be following 2.5.

That said, I'm a bit leery of spreading the validators amongst all the various component repositories:

- Each repo that has validators now has a dependency on zend-validator.
- Consumption is more difficult, as you now need to know what component packages define validators, vs. having them in one location, and developers will need to add configuration to the validator plugin manager in order to expose the validators from other packages.

The second can be solved, of course, by modifying the current "suggest" section of the `composer.json` to refer to the validators instead of the dependency; however, it still provides overhead in consuming the validators for the end-user, as they both need to install the additional package(s) and then configure the validators in the plugin manager.

We've seen already that having validators split between zend-i18n and zend-validator causes problems when folks are looking for common validators such as alpha, alnum, DateTime, IsFloat, and IsInt; they expect them to be in zend-validator, and have no idea that they might be elsewhere.

akrabat:
Also, as someone who uses Zend\Validator standalone, I don't want to bring in the whole world. If anything I'd like basic i18n ones to be in Zend\Validator.

[...]

PR #15, opened 2015-07-07
Update validators to allow instantiation without options

alexdenvir:
* Rather than throw an exception in the constructor, throw an exception in isValid
 * Updates Between, GreaterThan, IsInstanceOf, LessThan and Regex

### DIFF
--- a/src/Between.php
+++ b/src/Between.php
@@ -44,8 +44,8 @@ class Between extends AbstractValidator
      */
     protected $options = [
         'inclusive' => true,  // Whether to do inclusive comparisons, allowing equivalence to min and/or max
-        'min'       => 0,
-        'max'       => PHP_INT_MAX,
+        'min'       => null,
+        'max'       => null,
     ];
 
     /**
@@ -66,7 +66,13 @@ class Between extends AbstractValidator
         }
         if (!is_array($options)) {
             $options = func_get_args();
-            $temp['min'] = array_shift($options);
+
+            $temp = [];
+
+            if (!empty($options)) {
+                $temp['min'] = array_shift($options);
+            }
+
             if (!empty($options)) {
                 $temp['max'] = array_shift($options);
             }
@@ -76,12 +82,6 @@ class Between extends AbstractValidator
             }
 
             $options = $temp;
-        }
-
-        if (count($options) !== 2
-            && (!array_key_exists('min', $options) || !array_key_exists('max', $options))
-        ) {
-            throw new Exception\InvalidArgumentException("Missing option. 'min' and 'max' have to be given");
         }
 
         parent::__construct($options);
@@ -163,6 +163,10 @@ class Between extends AbstractValidator
     public function isValid($value)
     {
         $this->setValue($value);
+
+        if (is_null($this->getMin()) || is_null($this->getMax())) {
+            throw new Exception\InvalidArgumentException("Missing option. 'min' and 'max' have to be given");
+        }
 
         if ($this->getInclusive()) {
             if ($this->getMin() > $value || $value > $this->getMax()) {
--- a/src/Between.php
+++ b/src/Between.php
@@ -164,7 +164,7 @@ class Between extends AbstractValidator
     {
         $this->setValue($value);
 
-        if (is_null($this->getMin()) || is_null($this->getMax())) {
+        if ($this->getMin() === null || $this->getMax() === null) {
             throw new Exception\InvalidArgumentException("Missing option. 'min' and 'max' have to be given");
         }
 
--- a/test/StaticValidatorTest.php
+++ b/test/StaticValidatorTest.php
@@ -154,7 +154,6 @@ class StaticValidatorTest extends \PHPUnit_Framework_TestCase
 
     public function testExecuteValidWithParameters()
     {
-        $this->assertTrue(StaticValidator::execute(5, 'Between', [1, 10]));
         $this->assertTrue(StaticValidator::execute(5, 'Between', ['min' => 1, 'max' => 10]));
     }
 }
--- a/test/ValidatorChainTest.php
+++ b/test/ValidatorChainTest.php
@@ -120,7 +120,6 @@ class ValidatorChainTest extends \PHPUnit_Framework_TestCase
 
     public function testIsValidWithParameters()
     {
-        $this->assertTrue(StaticValidator::execute(5, 'Between', [1, 10]));
         $this->assertTrue(StaticValidator::execute(5, 'Between', ['min' => 1, 'max' => 10]));
     }
 
--- a/src/GreaterThan.php
+++ b/src/GreaterThan.php
@@ -139,7 +139,7 @@ class GreaterThan extends AbstractValidator
     {
         $this->setValue($value);
 
-        if (is_null($this->getMin())) {
+        if ($this->getMin() === null) {
             throw new Exception\InvalidArgumentException("Missing option 'min'");
         }
 
--- a/src/GreaterThan.php
+++ b/src/GreaterThan.php
@@ -64,7 +64,12 @@ class GreaterThan extends AbstractValidator
         }
         if (!is_array($options)) {
             $options = func_get_args();
-            $temp['min'] = array_shift($options);
+
+            $temp = [];
+
+            if (!empty($options)) {
+                $temp['min'] = array_shift($options);
+            }
 
             if (!empty($options)) {
                 $temp['inclusive'] = array_shift($options);
@@ -73,16 +78,9 @@ class GreaterThan extends AbstractValidator
             $options = $temp;
         }
 
-        if (!array_key_exists('min', $options)) {
-            throw new Exception\InvalidArgumentException("Missing option 'min'");
-        }
-
         if (!array_key_exists('inclusive', $options)) {
             $options['inclusive'] = false;
         }
-
-        $this->setMin($options['min'])
-             ->setInclusive($options['inclusive']);
 
         parent::__construct($options);
     }
@@ -141,13 +139,17 @@ class GreaterThan extends AbstractValidator
     {
         $this->setValue($value);
 
-        if ($this->inclusive) {
-            if ($this->min > $value) {
+        if (is_null($this->getMin())) {
+            throw new Exception\InvalidArgumentException("Missing option 'min'");
+        }
+
+        if ($this->getInclusive()) {
+            if ($this->getMin() > $value) {
                 $this->error(self::NOT_GREATER_INCLUSIVE);
                 return false;
             }
         } else {
-            if ($this->min >= $value) {
+            if ($this->getMin() >= $value) {
                 $this->error(self::NOT_GREATER);
                 return false;
             }
--- a/src/IsInstanceOf.php
+++ b/src/IsInstanceOf.php
@@ -97,7 +97,7 @@ class IsInstanceOf extends AbstractValidator
      */
     public function isValid($value)
     {
-        if (is_null($this->getClassName())) {
+        if ($this->getClassName() === null) {
             throw new Exception\InvalidArgumentException('Missing option "className"');
         }
 
--- a/src/IsInstanceOf.php
+++ b/src/IsInstanceOf.php
@@ -56,13 +56,12 @@ class IsInstanceOf extends AbstractValidator
             $options = func_get_args();
 
             $tmpOptions = [];
-            $tmpOptions['className'] = array_shift($options);
+
+            if (!empty($options)) {
+                $tmpOptions['className'] = array_shift($options);
+            }
 
             $options = $tmpOptions;
-        }
-
-        if (!array_key_exists('className', $options)) {
-            throw new Exception\InvalidArgumentException('Missing option "className"');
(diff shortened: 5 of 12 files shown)


alexdenvir:
This is an alternative solution to, and supercedes https://github.com/zendframework/zend-validator/pull/14

From the discussion there, it would be nicer to be able to instantiate validators without any options in the constructor, but still throw an exception when the validator **is used** is the required options have not been set

alexdenvir:
Pinging @gianarb @freax @manuakasam @Mezzle for comments on this as well, as they were involved in the original discussion on https://github.com/zendframework/zend-validator/pull/14

Mezzle:
Given the previous version required you to pass the params, not sure why it even had defaults.

manuakasam:
:+1: thanks for this contribution. I think this is a much better approach than what was started in #14 

I like this one presented here!

test/StaticValidatorTest.php
```diff
@@ -154,7 +154,6 @@ public function testPassingNullWhenSettingPluginManagerResetsPluginManager()
 
     public function testExecuteValidWithParameters()
     {
-        $this->assertTrue(StaticValidator::execute(5, 'Between', [1, 10]));
```
samsonasik: don't remove it, this should be still valid.

test/ValidatorChainTest.php
```diff
@@ -120,7 +120,6 @@ public function testStaticFactoryClassNotFound()
 
     public function testIsValidWithParameters()
     {
-        $this->assertTrue(StaticValidator::execute(5, 'Between', [1, 10]));
```
samsonasik: don't remove it, this should be still valid.

src/Between.php
```diff
@@ -164,6 +164,10 @@ public function isValid($value)
     {
         $this->setValue($value);
 
+        if (is_null($this->getMin()) || is_null($this->getMax())) {
```
freax: It's better to perform this kind of checks here and whenever like `$smth !== null`

samsonasik: don't use <code>is_null</code> for consistency, use <code>$this->getMin() === null</code> instead, see https://github.com/zendframework/zf2/pull/6129 and https://github.com/zendframework/zf2/pull/6643

test/StaticValidatorTest.php
alexdenvir: Actually, I'm not sure it ever was. It definitely isn't valid with my changes at least.

As far as I can tell, passing this array in to the constructor for Between results in the following set as options (in the master branch):
```php
protected $options = [
    // default options in master branch
    'inclusive' => true,
    'min' => 0,
    'max' => PHP_INT_MAX,
    // options created by parent:__construct
    0 => 1,
    1 => 10
];

Because the class has default options (in master), no error is encountered in this test because 5 **is** between 0 and PHP_INT_MAX

Comment (alexdenvir) [inline test/StaticValidatorTest.php]:
```
 % php -a                                                                           code/zend-validator (master) two
Interactive mode enabled

php > require_once "vendor/autoload.php";
php > $validator = new \Zend\Validator\Between([0, 10]);
php > die(var_dump($validator));
class Zend\Validator\Between#2 (5) {
  protected $messageTemplates =>
  array(2) {
    'notBetween' =>
    string(57) "The input is not between '%min%' and '%max%', inclusively"
    'notBetweenStrict' =>
    string(53) "The input is not strictly between '%min%' and '%max%'"
  }
  protected $messageVariables =>
  array(2) {
    'min' =>
    array(1) {
      'options' =>
      string(3) "min"
    }
    'max' =>
    array(1) {
      'options' =>
      string(3) "max"
    }
  }
  protected $options =>
  array(5) {
    'inclusive' =>
    bool(true)
    'min' =>
    int(0)
    'max' =>
    int(9223372036854775807)
    [0] =>
    int(0)
    [1] =>
    int(10)
  }
  protected $value =>
  NULL
  protected $abstractOptions =>
  array(7) {
    'messages' =>
    array(0) {
    }
    'messageTemplates' =>
    array(2) {
      'notBetween' =>
      string(57) "The input is not between '%min%' and '%max%', inclusively"
      'notBetweenStrict' =>
      string(53) "The input is not strictly between '%min%' and '%max%'"
    }
    'messageVariables' =>
    array(2) {
      'min' =>
      array(1) {
        ...
      }
      'max' =>
      array(1) {
        ...
      }
    }
    'translator' =>
    NULL
    'translatorTextDomain' =>
    NULL
    'translatorEnabled' =>
    bool(true)
    'valueObscured' =>
    bool(false)
  }
}
```

Comment (samsonasik) [inline test/StaticValidatorTest.php]: iirc, the behavior is, on StaticValidator::execute, the args is an array of parameter. at this case, args will be extracted. First parameter as min, and second as max. 

Warm regards,

Abdul Malik Ikhsan

Pada 8 Jul 2015, pukul 16.13, Alex Denvir <notifications@github.com> menulis:

Comment (alexdenvir) [inline test/StaticValidatorTest.php]: I agree, it probably should.
But I'm afraid it doesn't:
```
 % php -a                                                                                                                                                                                                   code/zend-validator (master) two
Interactive mode enabled

php > require_once "vendor/autoload.php";
php > var_dump(\Zend\Validator\StaticValidator::execute(5, 'Between', ['min' => 1, 'max' => 10]));
bool(true)
php > var_dump(\Zend\Validator\StaticValidator::execute(0, 'Between', ['min' => 1, 'max' => 10]));
bool(false)
php > var_dump(\Zend\Validator\StaticValidator::execute(100, 'Between', ['min' => 1, 'max' => 10]));
bool(false)
php > var_dump(\Zend\Validator\StaticValidator::execute(5, 'Between', [1, 10]));
bool(true)
php > var_dump(\Zend\Validator\StaticValidator::execute(0, 'Between', [1, 10]));
bool(true)
php > var_dump(\Zend\Validator\StaticValidator::execute(100, 'Between', [1, 10]));
bool(true)
```

Comment (samsonasik) [inline test/StaticValidatorTest.php]: Oops, sorry. It seems you're right ^^. How about kept it with pass 3 parameters [false, 1, 10] ? 
I am away from pc now ;)

Warm regards,

Abdul Malik Ikhsan

Pada 8 Jul 2015, pukul 18.00, Alex Denvir <notifications@github.com> menulis:

Comment (samsonasik) [inline test/StaticValidatorTest.php]: @alexdenvir I've just checked it in mine, It indeed strange:
```
php > var_dump( (new \Zend\Validator\Between(1, 10))->isValid(100) );
bool(false)

php > var_dump(\Zend\Validator\StaticValidator::execute(100, 'Between', [1, 10]));
bool(true)
```
hm...

File: src/Between.php (hunk 3 above)
Comment (Mezzle) [inline src/Between.php]: How is this consistent if there's 

```php
if (null === $foo->bar())
```
and
```php
if ($foo->bar() === null)
```

test/StaticValidatorTest.php
Mezzle: The second test here is actually better.

Originally it was testing for 5 being between 1 and 10, which would have always been true with the defaults.  It never tested for the negative case though.

alexdenvir: I have opened a separate issue to address this outside of this pull request: https://github.com/zendframework/zend-validator/issues/16

gianarb:
Sorry maybe I'm lost something, why an array?
```
new \Zend\Validator\Between([0, 10]);
```

test/StaticValidatorTest.php
samsonasik: @alexdenvir it seems the old one is an issue that need to be updated as args is an array, so it should be:

```php
var_dump(\Zend\Validator\StaticValidator::execute(100, 'Between', ['min' => 1, 'max' => 10]));
bool(false)

php > var_dump(\Zend\Validator\StaticValidator::execute(5, 'Between', ['min' => 1, 'max' => 10]));
bool(true)
```

So, it should can be updated with:

```
$this->assertTrue(StaticValidator::execute(5, 'Between', ['min' => 1, 'max' => 10]));
```

samsonasik:
@gianarb yeah, the old one on StaticValidator call seems an issue, and can be updated with:
```php
$this->assertTrue(StaticValidator::execute(5, 'Between', ['min' => 1, 'max' => 10]));
```

test/StaticValidatorTest.php
alexdenvir: That's exactly what the next line in the test is :wink:

samsonasik: cool ;). just aware of that, thanks ;)

src/Between.php
```diff
@@ -164,6 +164,10 @@ public function isValid($value)
     {
         $this->setValue($value);
 
+        if ($this->getMin() === null || $this->getMax() === null) {
+            throw new Exception\InvalidArgumentException("Missing option. 'min' and 'max' have to be given");
```
Ocramius: I disagree with this approach. What's the actual use-case for this change?

I'm basically just missing this bit: why do you want to instantiate a validator and keep it in an invalid state?

GeeH:
I don't think this brings anything of value to the table.

I actually would prefer you write your own implementations of these validators if you want defaults, I can't see that, for example, only passing in a max and not min to a between validator is improving anything.

GeeH:
To clarify, I think that ensuring that the min/max values are passed in via the options array is the correct way to use this component, and while it might be useful to override these values after construction, I don't see what this PR improves.

alexdenvir:
@GeeH - copy/pasted from a comment I made on the previous PR:

Ultimately the reason for me creating this PR is that I'm not a fan of being forced to pass an array of options in a constructor, as the values attached to array keys cannot be typehinted, there is no IDE autocomplete for the array keys themselves etc.

Compare calling:
```php
$validator = new Between(['min' => 5, 'max' => 10]);
```
to:
```php
$validator = new Between();
$validator->setMin(5)->setMax(10);
```
Any good IDE will autocomplete the function calls, while also telling you exactly what the function is doing and what value it is expecting, whether the value is typehinted etc.

Ocramius:
```php
$validator = new Between();
$validator->setMin(5)->setMax(10);
```

This is NOT good code.

`$validator` should be completely runtime-ready as soon as it's constructed, or else it's just bugs waiting to happen.

src/Between.php
alexdenvir: As explained in the reply I just made to @Geeh, it's not to let you instantiate a validator and keep it in an invalid state, but instead to allow you to avoid using the "array of options with arbitrary string keys" if you want to.

Related issue I opened https://github.com/zendframework/zend-validator/issues/16 - the current tests show an example of the Between validator being instantiated with invalid options, but not throwing an exception (because there are 2 values in the array), and the validator is then giving false positives because t falls back to the default values (0 and PHP_INT_MAX). This PR works around that, because if you pass in invalid options to the constructor (leaving the validator in an invalid state), the isValid function will throw the exception

Mezzle:
@Ocramius 
> This is NOT good code.

then surely the definition of the constructor should be

```php
public function __construct($min, $max)
{
    // ...
}
```

Rather than requiring what is basically a list of options, in a specific order, not an assosciative array as the documentation would suggest.

```
php > $foo = ['max' => 10, 'inclusive' => true, 'min' => 1];
php > $tmp['min'] = array_shift($foo);
php > var_dump($tmp);
array(1) {
  'min' =>
  int(10)
}
```

alexdenvir:
@Ocramius "This is NOT good code." - seems a bit harsh considering its a difference of opinion, but whatever.

As discussed in previous comments against this PR (and the previous), the Between validator has a default min of `0` and a default max of `PHP_INT_MAX`.
As it currently stands in master, you can instantiate the Between validator in the following (valid) ways:
```php
$validator1 = new Between(['min' => 5, 'max' => 10);
$validator2 = new Between(5, 10);
```
But you can also instantiate it in invalid ways and have no Exception thrown:
```php
$bad_validator = new Between(['minimum' => 5, 'maximum' => 10]);
```
In this example, the validator is not doing what is expected, you haven't passed a valid array of options, and isValid is going to produce false positives:
```php
$bad_validator->isValid(3);
// returns true
```

If the checks are done and Exceptions are thrown in `isValid` instead of the constructor, then the bad validator will handle this in an appropriate manner. This also opens up the option to instantiate validators with no options (which was my original intention, but its evolved from there). This pull request is not stopping anyone from continuing to use these validators as they have previously been used - you can still pass an array of valid options (or individual options through func_get_args magic) and have the validator work as expected

GeeH:
@alexdenvir - I'm sorry, I just don't agree. The validator **needs** a minimum and maximum value set before it can validate. That's a hard configuration dependency, and therefore these should be required at construction time.

weierophinney:
Just a note: Allowing an array of options to pass to the constructor allows for generalized factories for instantiation; without them, we'd need a factory per validator (or we'd need to define a `factory` method as part of the general interface). That said, I fully support the idea that setters should not be used here; validators should be configured once per instance, and remain stateless. How we get there, however, is for another story (*cough* #1 *cough*).

As such, I tend to agree with @GeeH and @Ocramius ; this patch is moving in a different direction than we ultimately intend to be, and for that reason, I'm closing it.

PR #17, opened 2015-07-11
Zend validator unit tests

rms230:
This pull request is more unit tests to improve the code coverage of zend-validator.

### DIFF
--- a/test/BitwiseTest.php
+++ b/test/BitwiseTest.php
@@ -226,5 +226,4 @@ class BitwiseTest extends \PHPUnit_Framework_TestCase
         $validator->setControl($control);
         $this->assertEquals($control, $validator->getControl());
     }
-    
 }
--- a/test/BitwiseTest.php
+++ b/test/BitwiseTest.php
@@ -177,4 +177,66 @@ class BitwiseTest extends \PHPUnit_Framework_TestCase
         $validator = new Bitwise(0x1, Bitwise::OP_AND, true);
         $this->assertTrue($validator->getStrict());
     }
+    
+    
+    public function testConstructWithArguments()
+    {
+    
+    	$control = 0x1;
+    	$operator = Bitwise::OP_AND;
+    	$strict = true;
+    
+    	$validator = new Bitwise($control, $operator, $strict);
+    	
+    	$this->assertEquals($control, $validator->getControl());
+    	$this->assertEquals($operator, $validator->getOperator());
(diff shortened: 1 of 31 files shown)


gianarb:
It's good IMO :+1:

Maks3w:
We are introducing a new way in the architecture of test that implies:

* test API methods 
* Use more data providers

The proposal made here do some assertions about non API methods which make the test less resilient 

My idea is take the test examples proposed here and include them in v3

gianarb:
This PR increase the value of this library because it cover a lot of possibile problem and it works now with this version..

I don't understand why this update can be a problem

test/BitwiseTest.php
```diff
@@ -177,4 +177,53 @@ public function testSetStrict()
...
+        $strict = true;
+        $validator = new Bitwise(0x1, Bitwise::OP_AND, $strict);
+        $this->assertEquals($strict, $validator->getStrict());
+    }
+
+    public function testIsValidWithInvalidOperator()
+    {
+        $validator = new Bitwise(0x1, 'or', false);
+        $expectedResult = false;
+        $this->assertEquals($expectedResult, $validator->isValid(0x2));
```
weierophinney: I'd argue you should use either `assertSame($expectedResult, $validator->isValid(...))` or, more simply, `assertFalse($validator->isValid(...))`. I can make that change on merge, however.

test/CallbackTest.php
```diff
@@ -116,4 +117,22 @@ public function optionsCallback($value)
         $this->assertContains('something', $args);
         return $args;
     }
+
+    public function testIsValidWithNoCallback()
```
weierophinney: Rename this to indicate what *behavior* you're testing: 'testIsValidRaisesExceptionInAbsenceOfCallback'. I can make that change on merge.

test/Db/AbstractDbTest.php
```diff
@@ -0,0 +1,90 @@
...
+ * @group      Zend_Validator
+ */
+class AbstractDbTest extends \PHPUnit_Framework_TestCase
+{
+
+    protected $validator;
+
+    public function setUp()
+    {
+        $this->validator = new ConcreteDbValidator([
```
weierophinney: A better approach would be to use mocking for the DB adapter. I'll see if I can work up an example.

weierophinney:
This patch improves the library *now*, and will help pave the way for v3. I'm scheduling it for the next bugfix release.

PR #18, opened 2015-07-13
Introducing GPSPoint Validator

zf2timo:
In this PR i want to introduce a new Validator, which validates GPS Coordinates. The requirements to the Validator are:
- Validate Decimal Degree Coordinates
- Validate Degree Minutes Second (DMS) Coordinates
- Validate minimum and maximum Values for feasible coordinates

Basic Unit Tests are included.
Please Review the code and  let me know your opinion.


src/GPSPoint.php
```diff
@@ -0,0 +1,131 @@
...
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Validator;
+
+final class GPSPoint extends AbstractValidator
```
freax: Why this is `final` class?

src/GPSPoint.php
```diff
@@ -0,0 +1,131 @@
...
+
+        return false;
+    }
+
+    /**
+     * @param string $value
+     * @param $maxBoundary
+     * @return bool
+     */
+    private function isValidCoordinate($value, $maxBoundary)
```
freax: Why this is `private` method?

weierophinney: @freax We've been gradually moving towards private visibility by default. This encourages good encapsulation, and moves the onus for justifying protected/public visibility to consumers; in other words, we need a very, very good reason to make something visible to extension writers.

Doing this eases our maintenance burden, because we can add, change, or remove private methods with little to no impact on end-users.

src/GPSPoint.php
```diff
@@ -0,0 +1,131 @@
...
+
+
+    /**
+     * @param string $value
+     * @return bool|string
+     */
+    private function convertValue($value)
+    {
+        $matches = [];
+        $result = preg_match_all('/(\d{1,3})°(\d{1,2})\'(\d{1,2}[\.\d]{0,6})"[NESW]/i', $value, $matches);
```
freax: - It can be `degrees only` or `degrees minutes` formats, like `120°W`, and this preg will fails.
- You can avoid `i` modifier by replace `[NESW]` to `[NESWnesw]`

src/GPSPoint.php
```diff
@@ -0,0 +1,131 @@
...
+        return $matches[1][0] + $matches[2][0]/60 + ((double)$matches[3][0])/3600;
+    }
+
+    /**
+     * @param string $value
+     * @return string
+     */
+    private function removeWhiteSpace($value)
+    {
+        return preg_replace('/\s/', '', $value);
```
freax: Can `\s` meant anything but `space`? If not, I think `str_replace` better.

src/GPSPoint.php
freax: Just look around a whole framework and you hardly notice a `private` methods. You need a very good reason to mark method as `private`. I think `protected` is better.

src/GPSPoint.php
weierophinney: `\s` is any whitespace character, and includes tab, CR, and EOL, among
others.

src/GPSPoint.php
weierophinney: Again, this is the direction we're *moving*. While we've defaulted to
protected in the *past*, we're changing practices based on the maintenance
issues the practice has created. (For your reference, I'm project lead; you
can take this as canonical.)

src/GPSPoint.php
freax: Yep. My point is that very unlikely there is need to check coordinate value in format like
120°

34'

23"
N

src/GPSPoint.php
manuakasam: @freax have a read at http://ocramius.github.io/blog/when-to-declare-classes-final/
This is a pretty easy to understand thing. When you declare classes final, there's no single reason to have protected functions as the class will never be extended. /cc @Ocramius

src/GPSPoint.php
zf2timo: See Link from @manuakasam in Comments below.

src/GPSPoint.php
zf2timo: - Ok, i will have a look, how this can be done.
- Why i should avoid the `i` modifier?

src/GPSPoint.php
zf2timo: A line break is very unlikely - that's right. But a `\t` is valid case and should not cause a validation error.

zf2timo:
With my commits, the Code Coverage decreased. Is there a way i can also cover the private Methods to increase the CC?

src/GPSPoint.php
freax: > When you declare classes final, there's no single reason to have protected functions as the class will never be extended.

Yes, it's pretty obvious. But my question is why need to forbid extend class from framework, especially from `validator` package.

froschdesign: > But my question is why need to forbid extend class from framework…

See again http://ocramius.github.io/blog/when-to-declare-classes-final/ (2. Encouraging composition)

freax: Well, if treat the whole article as the ultimate truth...

vaclavvanik:
In regard of naming conventions this should be renamed to GpsPointValidator

zf2timo:
I think the suffix validator is needless because the word is already in the Namespace. Also all other Validators did not have this suffix.
The CamelCase is acceptable.

vaclavvanik:
@zf2timo Sorry, I meant GpsPoint only.

src/GpsPoint.php
```diff
@@ -0,0 +1,131 @@
...
+     * @throws Exception\RuntimeException If validation of $value is impossible
+     */
+    public function isValid($value)
+    {
+        if (strpos($value, ',') === false) {
+            $this->error(GpsPoint::INCOMPLETE_COORDINATE, $value);
+            return false;
+        }
+
+        list($lat, $long) = explode(',', $value);
```
weierophinney: Shouldn't the `explode` have the third argument `2` to ensure no more than 2 values are returned?

weierophinney:
Merged to develop for release with 2.6.0

src/GpsPoint.php
zf2timo: @weierophinney Yes, it's cleaner. But should i make this change, because it was already merged.

weierophinney: I made the change during merge, iirc.

PR #23, opened 2015-07-20
Add isNotValid api

gianarb:
I don't know if it's a good idea but I see a lot of this implementation
```php
if (!$validator->isValid()) {
}
```
In my opinion this is difficult to read.
I know that a new api is an important upgrade but this is very easy to maintain and the code is more readable!
```php
if ($validator->isNotValid()) {
}
```
In this PR I have not added it into the Interface because it would be a big **BC Break**.. :)


Maks3w:
I don't disagree with this.

weierophinney:
I *do* disagree with this. `isValid()` is a boolean method, and this is what the boolean operator (i.e., `!`) is for. If you're testing for something being invalid, you test `! $validator->isValid($data)`.

More methods === more to support, and more to learn.

PR #25, opened 2015-07-21
Date refactorting to 5.5

gianarb:


Ocramius:
Tests please!

gianarb:
thanks @Ocramius can you help me to write a good tests to coverage this feature?
 it's already coverage IMO

Ocramius:
> it's already coverage IMO

The particular path is not followed. You'd simply `$this->getMock(DateTimeInterface::class)` and work with that

gianarb:
Why not?
https://github.com/zendframework/zend-validator/blob/master/test/DateTest.php#L85

DateTime implement DateTimeinterface

Ocramius:
@gianarb but you are not testing that any implementation of `DateTimeInterface` passes the test.

samsonasik:
@gianarb thanks for this ;)

gianarb:
Little problem
http://3v4l.org/8GvgO

https://github.com/php/php-src/blob/4a2e40bb861bc3cf5fb6863e57486ed60316e97c/ext/date/php_date.c#L1961

Thanks @fntlnz :P

Ocramius:
@gianarb as it stands, you'd probably have to test with `DateTime` and `DateTimeImmutable` then. I raised the question at http://news.php.net/php.internals/87217

gianarb:
in this moment we have already wrote tests with DateTime, DateTime Immutable and stdClass

samsonasik:
@Ocramius tests updated by @gianarb ;)

Maks3w:
This is pending of RFC with a proposal of Interface removal. This will take days or weeks.

src/Date.php
```diff
@@ -127,8 +127,7 @@ public function isValid($value)
      */
     protected function convertToDateTime($param, $addErrors = true)
     {
-        // @TODO: when minimum dependency will be PHP 5.5, we can only keep check against DateTimeInterface
-        if ($param instanceof DateTime || $param instanceof DateTimeInterface) {
+        if ($param instanceof DateTimeInterface) {
```
Ocramius: As per discussion with the internal folks (sigh), this should probably be:

```php
if ($param instanceof DateTime || $param instanceof DateTimeImmutable) {
```

According to internal folks, `DateTimeInterface` is not to be used.

samsonasik:
@Ocramius , @gianarb has updated to instanceof DateTimeImmutable ;)

samsonasik:
@Ocramius @gianarb I think the PR title can be updated

Issue #16, opened 2015-07-09
StaticValidator test is incorrect (false positive)


alexdenvir:
Opening an issue as a result of findings in my pull request https://github.com/zendframework/zend-validator/pull/15. Here the issue has been removed, but not explicitly addressed.

There is a test in StaticValidatorTest (and ValidatorChainTest) which test the StaticValidator and Between:

```php
    public function testExecuteValidWithParameters()
    {
        $this->assertTrue(StaticValidator::execute(5, 'Between', [1, 10]));
        $this->assertTrue(StaticValidator::execute(5, 'Between', ['min' => 1, 'max' => 10]));
    }
```
The first assertion **implies** (or at least creates some confusion, as evidenced in the pull request) that you can pass an array of options in, and they are passed to the actual validator as individual parameters. This is definitely not the case.
In this example, the Between validator will end up with an option called '0', with a value of '1', and an option called '1', with a value of '10'. The test currently passes, because the Between validator has a default value for min and max (0 and PHP_INT_MAX respectively).

To show more detail:

```
 % php -a
Interactive mode enabled

php > require_once "vendor/autoload.php";
php > var_dump(\Zend\Validator\StaticValidator::execute(5, 'Between', ['min' => 1, 'max' => 10]));
bool(true)
php > var_dump(\Zend\Validator\StaticValidator::execute(0, 'Between', ['min' => 1, 'max' => 10]));
bool(false)
php > var_dump(\Zend\Validator\StaticValidator::execute(100, 'Between', ['min' => 1, 'max' => 10]));
bool(false)
php > var_dump(\Zend\Validator\StaticValidator::execute(5, 'Between', [1, 10]));
bool(true)
php > var_dump(\Zend\Validator\StaticValidator::execute(0, 'Between', [1, 10]));
bool(true)
php > var_dump(\Zend\Validator\StaticValidator::execute(100, 'Between', [1, 10]));
bool(true)
php > $validator = new \Zend\Validator\Between([1, 10]);
php > var_dump($validator);
class Zend\Validator\Between#2 (5) {
  protected $messageTemplates =>
  array(2) {
    'notBetween' =>
    string(57) "The input is not between '%min%' and '%max%', inclusively"
    'notBetweenStrict' =>
    string(53) "The input is not strictly between '%min%' and '%max%'"
  }
  protected $messageVariables =>
  array(2) {
    'min' =>
    array(1) {
      'options' =>
      string(3) "min"
    }
    'max' =>
[...]

PR #26, opened 2015-07-21
Fix StaticValidator::execute tests and behavior

weierophinney:
This is an attempt to address #16.

Previously, the tests were incorrectly passing a non-associative array of options to `execute()`, which implied that they would be passed as constructor arguments, which was not the case. In this particular case, there was a false positive, as the value being validated was valid for the default values present in the instance.

This patch removes that assertion, and splits the test into two, one for validating properly formed calls to the method with options as an associative array, and one for calling it with non-associative options. In the second, the test asserts that an `InvalidArgumentException` is expected.

The patch is a potential BC break. However, passing a non-associative array of arguments had unexpected behavior previously, and relying on that behavior would result in application bugs. The change to raise an exception ensures consistent behavior.


src/StaticValidator.php
```diff
@@ -52,6 +52,12 @@ public static function getPluginManager()
      */
     public static function execute($value, $classBaseName, array $args = [])
     {
+        if ($args && array_values($args) === $args) {
```
Maks3w: Should this check to be done in the PluginManager?

src/StaticValidator.php
```diff
@@ -52,6 +52,12 @@ public static function getPluginManager()
      */
```
Maks3w: Document `@throws`

src/StaticValidator.php
Maks3w: Forgive this

Maks3w:
I see three cases here:

* Constructor uses `func_get_args`
* $args is used for constructor arguments so could be a bidimensional array passing the options as first argument
`[ [ options ] ]
* $args is the array options accidentaly used as constructor arguments list

Because argument name is `$args` which means `arguments list` I think the current implementation is the correct. The intention of provide a constructor argument list.

Provide a constructor argument list allow to instantiate constructors with different constructor signatures and not only the ones based in AbstractValidator.

My opinion is not merge and won't fix.

Maks3w:
I've forwarded my comment to the issue for to maintain a single thread.

weierophinney:
> Constructor uses func_get_args

This one is a moot point; the plugin manager *always* passes arguments (options) as an array, never using variadics.

> $args is used for constructor arguments so could be a bidimensional array passing the options as first argument `[ [ options ] ]`

This does not occur currently.

> $args is the array options accidentaly used as constructor arguments list

`$args` is passed as the `$options` argument to the plugin manager instance, and `$options`, if present, is always passed as the sole argument to the plugin constructor. The specified use case when we designed this was precisely for passing an associative array (or array-like object) of options to the plugin, to allow setting its initial state.

Which completely invalidates this assertion:

> Because argument name is $args which means arguments list I think the current implementation is the correct. The intention of provide a constructor argument list.

That was never the intention. If it were, **I would have coded the `AbstractPluginManager` differently**. This is simply a case of (a) misnamed argument, and (b) incorrect initial tests. Somebody essentially discovered that we were not covering an edge case properly.

With that knowledge, please re-review.

src/StaticValidator.php
Maks3w: As per discussion on IRC

* $args param need to improve his description
* $args argument should be renamed

weierophinney: Latest commit:

- documents `@throws`
- Renames `$args` to `$options`
- Better documents `$options`

src/StaticValidator.php
```diff
@@ -47,14 +47,22 @@ public static function getPluginManager()
     /**
      * @param  mixed    $value
      * @param  string   $classBaseName
-     * @param  array    $args          OPTIONAL
+     * @param  array    $optons OPTIONAL options to pass to the validator constructor;
```
Maks3w: * typo s/optons/options
* Include "as first argument". "associative array of options for to inject in the first argument of the validator constructor"

test/StaticValidatorTest.php
```diff
@@ -152,9 +152,38 @@ public function testPassingNullWhenSettingPluginManagerResetsPluginManager()
...
+            'invalid-negative-range' => [5, 'Between', ['min' => -10, 'max' => -1], 'assertFalse'],
+        ];
+    }
+
+    /**
+     * @dataProvider parameterizedData
+     */
+    public function testExecuteValidWithParameters($value, $validator, $options, $assertion)
+    {
+        $this->$assertion(StaticValidator::execute($value, $validator, $options));
```
Maks3w: Lately I prefer asser the type first and then the value

asserInternalType(boolean, return) 
assertEquals(expected, return) 

The first ensure the type is a documented return type and also force exprcted to be the same type
The second one is the classic discrete comparison of values

This format has 2 goodnes:

* Keep clen the data provider from phpunit internals (method names) 
* Keep a consistent test format with a more clear error messages

src/StaticValidator.php
```diff
@@ -47,14 +47,22 @@ public static function getPluginManager()
...
-     * @param  array    $args          OPTIONAL
+     * @param  array    $options OPTIONAL associative array of options to pass as
+     *     the sole argument to the validator constructor.
      * @return bool
+     * @throws Exception\InvalidArgumentException for an invalid $options argument.
      */
-    public static function execute($value, $classBaseName, array $args = [])
+    public static function execute($value, $classBaseName, array $options = [])
     {
+        if ($options && array_values($options) === $options) {
```
Maks3w: test `0 === key($options)` could be enough?

weierophinney: The above is a standard, performant, robust solution that is less likely to obtain false positives. (It is the solution used in `Zend\Stdlib\ArrayUtils::isList()`.)

test/StaticValidatorTest.php
weierophinney: In this case, it can be simplified even further: `assertSame($expected, $return)`, which is the approach I'm about to push.

Maks3w: The problem with assersame is anyone can put a string in the provider.

Split in two make more clear 1:1 relations with method signature and avoid put false positive values in provider. 

This is motivated in cases where the provider is implemented in child classes while the test method is not

weierophinney: The contract for validators is to return boolean. Adding the assertion on
type is superfluous.

On Wed, Jul 22, 2015 at 8:01 AM, Maks3w <notifications@github.com> wrote:

> In test/StaticValidatorTest.php
> <https://github.com/zendframework/zend-validator/pull/26#discussion_r35209413>
> :
>
> > -        $this->assertTrue(StaticValidator::execute(5, 'Between', [1, 10]));
> > -        $this->assertTrue(StaticValidator::execute(5, 'Between', ['min' => 1, 'max' => 10]));
> > +        return [
> > +            'valid-positive-range'   => [5, 'Between', ['min' => 1, 'max' => 10], 'assertTrue'],
> > +            'valid-negative-range'   => [-5, 'Between', ['min' => -10, 'max' => -1], 'assertTrue'],
> > +            'invalid-positive-range' => [-5, 'Between', ['min' => 1, 'max' => 10], 'assertFalse'],
> > +            'invalid-negative-range' => [5, 'Between', ['min' => -10, 'max' => -1], 'assertFalse'],
> > +        ];
> > +    }
> > +
> > +    /**
> > +     * @dataProvider parameterizedData
> > +     */
> > +    public function testExecuteValidWithParameters($value, $validator, $options, $assertion)
> > +    {
> > +        $this->$assertion(StaticValidator::execute($value, $validator, $options));
>
> The problem with assersame is anyone can put a string in the provider.
>
> Split in two make more clear 1:1 relations with method signature and avoid
> put false positive values in provider.
>
> This is motivated in cases where the provider is implemented in child
> classes while the test method is not
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zendframework/zend-validator/pull/26/files#r35209413>.
>



-- 
Matthew Weier O'Phinney
matthew@weierophinney.net
https://mwop.net/

PR #29, opened 2015-07-22
Continue improving unit tests

weierophinney:
This pull request builds on #17:

- It renames new test method to better indicate the behavior tested.
- It provides data providers for the new test methods where warranted (and expands on those already provided)
- It adds data providers to existing test methods in several instances.
- It fixes logic errors detected when adding the data providers.

Assuming it passes on Travis, I plan to merge as this is essentially a review of and improvements on #17.


weierophinney:
Tests now pass on PHP 7, and the changes remove all test **errors** from HHVM. Three failures are still present on HHVM, and appear to be discrepancies with `getmxrr` on that platform.

PR #30, opened 2015-07-23
Automatize TLDs update

### LINKED ISSUE #3: Automatize update of valid TLDs
Actually Hostname validator depends of a regular execution of bin/update_hostname_validator.php for to host an up to date list of IANA TLDs

This should be automatized and even the list could be published in a different package for to allow more regular updates for LTS versions and newer.

Discussion: https://github.com/zendframework/zend-validator/pull/2

fntlnz:
Resolves #3 
I edited the original script to do the following things:

- If the `--check-only` option is passed, just check if TLDs are up to date and if not exit with 1
- Otherwise update the TLDs adding a reference to the Iana version number.

A first fail on travis is expected in order to test if it's working as expected. (Should fail because TLDs are not up to date)


bin/update_hostname_validator.php
```diff
@@ -25,35 +26,33 @@
...
+    exit(0);
 }
 
-$newFileContent = [];  // new file content
-$insertDone     = false;    // becomes 'true' when we find start of $validTlds declaration
-$insertFinish   = false;    // becomes 'true' when we find end of $validTlds declaration
 foreach (file(ZF2_HOSTNAME_VALIDATOR_FILE) as $line) {
+    // Replace old version number with new one
+    if (preg_match('/\*\s+Version\s+\d+/', $line, $matches)) {
+        $newFileContent[] = sprintf('     * Version %s%s', $ianaVersion, PHP_EOL);
```
Maks3w: no eol always \n

src/Hostname.php
```diff
@@ -67,6 +67,7 @@ class Hostname extends AbstractValidator
 
     /**
      * Array of valid top-level-domains
+     * Version 2015072300
```
Maks3w: We could use here IanaVersion so the word becomes unique and also express this is not our release version

.travis.yml
```diff
@@ -41,6 +41,7 @@ before_install:
 
 install:
   - travis_retry composer install --no-interaction --ignore-platform-reqs
+  - php bin/update_hostname_validator.php --check-only
```
Maks3w: * move this to the last script command
* make the command to execute only if CHECK_HOSTNAME env var is set. This will set only on php 5.5

bin/update_hostname_validator.php
```diff
@@ -95,10 +94,73 @@
...
+ * @param string $string
+ * @return string
+ * @throws Exception
+ */
+function getVersionFromString($string)
+{
+    $matches = [];
+
+    if (!preg_match('/Version\s+((\d+)?)/', $string, $matches)) {
+        throw new \Exception(sprintf('Error: cannot get last update date%s', PHP_EOL));
```
Maks3w: exception message shouldn't have eol

bin/update_hostname_validator.php
fntlnz: Ok

src/Hostname.php
fntlnz: Ok it's definitely better

bin/update_hostname_validator.php
fntlnz: nvm, I don't know why I did this

.travis.yml
fntlnz: Ok, the variable name shouldn't be `EXECUTE_HOSTNAME_CHECK` ?

Maks3w: ok

fntlnz: Another thing, I preferred to run the check right after install because I thought that one could prefer to see the build fail earlier as possible but this may vary on what one is looking for.

Maks3w: Block the execution of normal scripts. This is there is no way of to know if any other PR in the queue is well written or not

Maks3w: Put this

`if [[ $EXECUTE_HOSTNAME_CHECK  == "true" && $TRAVIS_PULL_REQUEST == "false" ]]; then bin/update_hostname_validator.php --check-only; fi`

fntlnz: Ok

fntlnz:
@Maks3w it's ok now?

Maks3w:
For me it's ok. I'll leave the final decission to @weierophinney since he wrotes that script.

With the proposed settings Travis-CI may fail only after merge a PR or rerun a build

fntlnz:
@weierophinney Does this looks good to you?

fntlnz:
@weierophinney @Maks3w Any problems with this?

weierophinney:
Merged with 5e4e671

PR #31, opened 2015-07-24
Fixed unused method

gianarb:
`errorHandlerIgnore` seems unused..
Can you help me to check if is it true?


fntlnz:
I think it should be done also [here](https://github.com/gianarb/zend-validator/blob/feature/refactoring-unused-method/test/ValidatorChainTest.php#L250)

PR #33, opened 2015-08-06
WIP Validator\Date with format 'd' return true with data 6

gianarb:
I wrote this test because 'd' is day of week with 2 digits but for Validator\Date 6 is valid


svycka:
would be nice if https://github.com/zendframework/zf2/pull/6407 would be fixed too.

```php
$validator = new Date(['format' => 'Y-m-d']);
$this->assertFalse($validator->isValid(123)); // fails
```

test/DateTest.php
```diff
@@ -118,6 +119,12 @@ public function testGetMessages()
         $this->assertEquals([], $this->validator->getMessages());
     }
 
+    public function testFormatDayTwoDigits()
+    {
+        $validator = new Date(['format' => 'd']);
+        $this->assertFalse($validator->isValid("6"));
```
Maks3w: Append validator failure message as second argument of assertFalse

gianarb: @Maks3w the problem is the result of validation not the failure message

Maks3w: @gianarb I know but is better if PHPUnit fail with a better description message rather true is not equal to false :)

Maks3w:
@gianarb Test case should be part of `datesDataProvider`

weierophinney:
Pinging @gianarb — do you have time to incorporate the feedback?

gianarb:
I do it now

weierophinney:
@gianarb Thanks.

Now the question is: is this a problem with the validator, or with `DateTime`...

weierophinney:
I'm still stumped by this, but now for a different reason: Are you asserting that given a format `d`, meaning a _day_, and a value of `'6'`, this is invalid because it is _not_ zero-padded? I'm not entirely sure I agree that this would be problematic behavior; I could see a case for any single digit (minus zero itself) being valid.

Can you clarify, please?

gianarb:
I can't because I don't remember anymore sorry.

PR #34, opened 2015-08-18
Update AbstractDb.php

andrey-mokhov:
When RecordExists and NoRecordExists validators are used two and more times, system throws exception:
SQLSTATE[HY093]: Invalid parameter number: :where1

This error is related with creation query string with parameters where2, where3, etc with every calling prepareStatementForSqlObject method.
This solution lets use RecordExists and NoRecordExists validators two and more times.


Maks3w:
Need a test

src/Db/AbstractDb.php
```diff
@@ -43,6 +44,11 @@
      * @var Select
      */
     protected $select;
+    
+    /**
+     * @var StatementInterface
+     */
+    protected $statement;
```
Maks3w: change to private

src/Db/AbstractDb.php
```diff
@@ -43,6 +44,11 @@
      * @var Select
      */
     protected $select;
+    
+    /**
+     * @var StatementInterface
```
Maks3w: StreamInterface|null

src/Db/AbstractDb.php
```diff
@@ -190,6 +196,7 @@ public function setExclude($exclude)
     {
         $this->exclude = $exclude;
         $this->select  = null;
+        $this->statement = null;
```
Maks3w: Looks like is needed to propose a reset() method for this repetition.

src/Db/AbstractDb.php
```diff
@@ -318,13 +330,27 @@ public function getSelect()
...
     }
+    
+    /**
+     * Prepares statement for Sql Object
+     * 
+     * @return StatementInterface
+     */
+    protected function getStatement()
+    {
+        if ($this->statement instanceof StatementInterface) {
```
Maks3w: Invert the logic here. If NOT then create and assign.

weierophinney:
@andrey-mokhov We cannot merge this change without tests; will you have time to implement them?

weierophinney:
Closing due to lack of updates.

PR #35, opened 2015-08-26
[NotEmpty] 0.0 should be treated as non empty by default

Maks3w:
Following the reasoning of allow `0` as non empty then `0.0` should be too.

http://framework.zend.com/issues/browse/ZF-6708

This makes the input filter to have a different behavior for 0.0


svycka:
related: https://github.com/zendframework/zf2/issues/6638

PR #36, opened 2015-08-26
Replace bit mask hex representation with binary one

Maks3w:
I think it's more visual and better understanding to see the 0/1 matrix.


weierophinney:
This isn't a BC break, just a change in internal representation; we can merge for 2.5.3.

PR #44, opened 2015-11-06
Improved the english on the greater than or equal to message

chateaux:
Added the correct english for the message.

### DIFF
--- a/src/GreaterThan.php
+++ b/src/GreaterThan.php
@@ -24,7 +24,7 @@ class GreaterThan extends AbstractValidator
      */
     protected $messageTemplates = [
         self::NOT_GREATER => "The input is not greater than '%min%'",
-        self::NOT_GREATER_INCLUSIVE => "The input is not greater or equal than '%min%'"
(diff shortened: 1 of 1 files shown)


chateaux:
Not quite sure what failed, but the gist of the change is to improve the readability of the text. I have kept it short, a more accurate version would be: The input is neither greater than nor equal to '%s'

src/GreaterThan.php
```diff
@@ -24,7 +24,7 @@ class GreaterThan extends AbstractValidator
      */
     protected $messageTemplates = [
         self::NOT_GREATER => "The input is not greater than '%min%'",
-        self::NOT_GREATER_INCLUSIVE => "The input is not greater or equal than '%min%'"
+        self::NOT_GREATER_INCLUSIVE => "The input is not greater or equal to '%min%'"
```
weierophinney: Ideally, this should read `greater than or equal to`. I can make that change on merge.

Test failures are because a test was looking for an exact exception string.

PR #45, opened 2015-11-06
fixed incorrect aliases

svycka:
more on this https://github.com/zendframework/zend-servicemanager/issues/50

### DIFF
--- a/src/ValidatorPluginManager.php
+++ b/src/ValidatorPluginManager.php
@@ -20,8 +20,8 @@ class ValidatorPluginManager extends AbstractPluginManager
      * @var array
      */
     protected $aliases = [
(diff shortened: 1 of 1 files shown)


src/ValidatorPluginManager.php
```diff
@@ -20,8 +20,8 @@ class ValidatorPluginManager extends AbstractPluginManager
      * @var array
      */
     protected $aliases = [
-        'Zend\I18n\Validator\Float'=> 'Zend\I18n\Validator\IsFloat',
-        'Zend\I18n\Validator\Int'  => 'Zend\I18n\Validator\IsInt',
+        'zendi18nvalidatorfloat'=> 'isfloat',
+        'zendi18nvalidatorint'  => 'isint',
```
weierophinney: Both should be present, tbh. I'll update on merge.

PR #46, opened 2015-12-07
Let AbstractDb implement AdapterAwareInterface

Keops92:


src/Db/AbstractDb.php
```diff
@@ -23,6 +24,8 @@
  */
 abstract class AbstractDb extends AbstractValidator
 {
+    use AdapterAwareTrait;
```
weierophinney: This is of no use unless you also implement `AdapterAwareInterface`. I'll add that when merging.

weierophinney:
Merged to develop for release with 2.6.0.

PR #49, opened 2016-01-07
Code refactor with the new servicemanager of ZF3

ezimuel:
This PR includes the code refactor with the new zend-servicemanager of ZF3.

### DIFF
--- a/src/StaticValidator.php
+++ b/src/StaticValidator.php
@@ -9,6 +9,8 @@
 
 namespace Zend\Validator;
 
+use Zend\ServiceManager\ServiceManager;
+
 class StaticValidator
 {
     /**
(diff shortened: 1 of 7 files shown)


src/StaticValidator.php
```diff
@@ -39,7 +41,7 @@ public static function setPluginManager(ValidatorPluginManager $plugins = null)
     public static function getPluginManager()
     {
         if (null === static::$plugins) {
-            static::setPluginManager(new ValidatorPluginManager());
+            static::setPluginManager(new ValidatorPluginManager(new ServiceManager));
```
Maks3w: This is a very common pattern repeated across all SM v3 PRs. May the SM should be injected by default?

weierophinney: @Maks3w note the context here: static usage. By very definition, no instance is available.

Within the MVC and the various plugin manager factories, the SM *will* be injected by default; that's the new design. (Instead of passing it to `setServiceLocator()`, it becomes a constructor argument, and becomes the "creation context", which means that it is that instance, not the plugin manager, passed to factories.)

weierophinney:
I'm marking this as 3.0 temporarily. Once zendframework/zend-servicemanager#59 is finalized and merged, we can likely do some modifications to make this a forwards-compat update for the next minor version.

PR #50, opened 2016-01-11
Fixed the initializer for the validator plugin manager

ezimuel:
This PR fixes the behaviour of the new PluginManager of zend-servicemanager due to the change reported [here](https://github.com/zendframework/zend-servicemanager/pull/59).


weierophinney:
Closing in favor of #51, which incorporates the commits from this patch.

PR #52, opened 2016-01-17
Fixes $options = null check when use Zend\Valdator\Barcode

samsonasik:
the constructor has optional `$options = null` but no check for it, this PR fixes it.

### DIFF
--- a/src/Barcode.php
+++ b/src/Barcode.php
@@ -48,6 +48,10 @@ class Barcode extends AbstractValidator
      */
     public function __construct($options = null)
     {
+        if ($options === null) {
+            $options = [];
+        }
+
         if (!is_array($options) && !($options instanceof Traversable)) {
             $options = ['adapter' => $options];
         }
--- a/test/BarcodeTest.php
+++ b/test/BarcodeTest.php
@@ -18,6 +18,22 @@ use Zend\Validator\Barcode;
  */
 class BarcodeTest extends \PHPUnit_Framework_TestCase
 {
+    public function provideBarcodeConstructor()
+    {
+        return [
+            [null, Barcode\Ean13::class],
+            [[], Barcode\Ean13::class],
+        ];
+    }
+    /**
+     * @dataProvider provideBarcodeConstructor
+     */
+    public function testBarcodeConstructor($options, $expectedInstance)
+    {
+        $barcode = new Barcode($options);
+        $this->assertInstanceOf($expectedInstance, $barcode->getAdapter());
+    }
+
     public function testNoneExisting()
     {
         $this->setExpectedException('Zend\Validator\Exception\InvalidArgumentException', 'not found');


Maks3w:
Next `if` initialize the adapter value as null and getAdatper has a check to for lazy initialization.

Can you paste the results of the added test without "the fix"?

test/BarcodeTest.php
```diff
@@ -18,6 +18,22 @@
...
+    {
+        return [
+            [null],
+            [[]],
+        ];
+    }
+    /**
+     * @dataProvider provideBarcodeConstructor
+     */
+    public function testBarcodeConstructorIsNullOrArrayButNoOptionsKey($options)
```
Maks3w: Don't detail constructor types in the function name, this is already provided by the provider.

src/Barcode.php
```diff
@@ -48,6 +48,10 @@ class Barcode extends AbstractValidator
      */
     public function __construct($options = null)
     {
+        if ($options === null) {
+            $options['options']['adapter'] = $this->getAdapter();
```
Maks3w: Breaks the lazy load of getAdapter

test/BarcodeTest.php
```diff
@@ -18,6 +18,22 @@
...
+    {
+        $barcode = new Barcode($options);
+        $this->assertInstanceOf('Zend\Validator\Barcode\Ean13', $barcode->getAdapter());
```
Maks3w: The expected instance of could be provided by the provider.

samsonasik:
without the fix, it will got error:
```
    Adapter NULL does not implement Zend\Validator\Barcode\AdapterInterface
```

src/Barcode.php
samsonasik: could you incorporate? should it be like this? :
```php
if ($options === null) {
       $options = [];
}
```

test/BarcodeTest.php
samsonasik: updated

test/BarcodeTest.php
samsonasik: updated.

Maks3w:
Then the source of the issue is:

1. `getAdapter` try to lazy initializate the adapter if this one does not exists but it access to the array key `adapter`
2. parent constructor calls `setAdapter` with null value and setAdapter check if the argument is a valid instance.

For me the easiest solution is allow `null` on `setAdapter`

samsonasik:
@Maks3w I've updated `setAdapter()` to allow null

src/Barcode.php
Maks3w: Sounds to be a valid solution.

Maks3w:
Your option of set an empty array was better

samsonasik:
if use ste empty array, it will got:
```
Zend\Validator\Exception\InvalidArgumentException: Adapter array does not implement Zend\Validator\Barcode\AdapterInterface
```

Maks3w:
With this? https://github.com/zendframework/zend-validator/pull/52#discussion_r49945163

samsonasik:
@Maks3w oops, missed that, I've just updated in https://github.com/zendframework/zend-validator/commit/2fd04e5e941f78435cfd574b48083964ebfcaabe

samsonasik:
@Maks3w should be ok now :+1:

Maks3w:
Rebase and go

Maks3w:
*squash

samsonasik:
@Maks3w squashed.

samsonasik:
@Maks3w merge-able ?

PR #56, opened 2016-02-03
escaped "." in regex for ipv4 Validation

BlckStar:
"." was not literal dot but any-char , IP "11111111111" was valid

### DIFF
--- a/src/Ip.php
+++ b/src/Ip.php
@@ -97,15 +97,15 @@ class Ip extends AbstractValidator
      */
     protected function validateIPv4($value)
     {
-        if (preg_match('/^([01]{8}.){3}[01]{8}\z/i', $value)) {
+	if (preg_match('/^([01]{8}\.){3}[01]{8}\z/i', $value)) {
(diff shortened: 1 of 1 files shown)


weierophinney:
We need unit tests to validate the change; can you please provide some, @BlckStar ?

weierophinney:
I've written unit tests. Interestingly, while the regex is incorrect, that fact does not result in false positives. The reason is because we then call `ip2long()` on the resultant value, which ends up invalidating the value. `111111111111` ends up invalid under the current code.

I'll still merge the change, as we *should* be escaping the `.` to ensure it's matched literally.

BlckStar:
Okay thanks, was stumbling over this. Sry i am kind of new at github and doing a pull request for the first time. But unit test sounds good. I will remember this.


Von Outlook Mobile gesendet






On Wed, Feb 17, 2016 at 9:08 AM -0800, "weierophinney" <notifications@github.com> wrote:












Merged #56.

PR #58, opened 2016-02-27
Uuid Validator

gsomoza:
I realise there's an open road to ZF3 PR but here's what I'd consider some low-hanging fruit in case you're planning any other 2.x releases. 

NOTE: I'd be happy to upgrade this to the new format once the new API for validators is stabilised in that other PR. So if you don't think you'll release this in a new 2.x release let me know and I'll rebase the work against the ZF3 WIP branch.

### DIFF
--- a/test/UuidTest.php
+++ b/test/UuidTest.php
@@ -1,0 +1,99 @@
+<?php
+
+/**
+ * Zend Framework (http://framework.zend.com/).
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ *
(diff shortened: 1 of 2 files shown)


test/UuidTest.php
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /**
+     * providesValidUuids.
+     *
+     * @return array
+     */
+    public function validUuidProvider()
+    {
+        return array(
```
samsonasik: can use short array syntax `[]`

gsomoza: Done!

weierophinney:
Thanks, @gsomoza !

gsomoza:
It's nothing! Happy to help a little bit :)

PR #61, opened 2016-03-08
Update EmailAddress.php

krakadin:
Currently, the regex is validating email addresses with the dot at the end of the local part and also multiple consecutive dots so email addresses such local.part.@domain.ext is valid and so is local...part@domain.ext. See RFC 5321 for allowed dot-atom formats. Proposed fix contains only removal of a + (plus) sign from the expression "if (preg_match('/^[' . $atext . ']+(\x2e[' . $atext . ']+)*$/', $this->idnToAscii($this->localPart))) {", right after \x2e.


samsonasik:
needs tests

Maks3w:
Close due inactivity

PR #65, opened 2016-04-18
Fix the module initializer

weierophinney:
Initializers actually receive the module manager itself, not the module event. As such, we have to pull the event from the module manager instance.

### DIFF
--- a/src/Module.php
+++ b/src/Module.php
@@ -24,11 +24,12 @@ class Module
     /**
      * Register a specification for the ValidatorManager with the ServiceListener.
      *
-     * @param \Zend\ModuleManager\ModuleEvent
+     * @param \Zend\ModuleManager\ModuleManager $moduleManager
      * @return void
      */
-    public function init($event)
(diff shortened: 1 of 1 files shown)


src/Module.php
```diff
@@ -24,11 +24,12 @@ public function getConfig()
     /**
      * Register a specification for the ValidatorManager with the ServiceListener.
      *
-     * @param \Zend\ModuleManager\ModuleEvent
+     * @param \Zend\ModuleManager\ModuleManager $moduleManager
      * @return void
      */
-    public function init($event)
+    public function init($moduleManager)
```
Maks3w: Is it possible to typehint the parameter?

weierophinney: Potentially. We've typically not done it, in order to prevent the inevitable "zend-modulemanager should be listed as a package requirement" issues. For some reason, not using a typehint doesn't trigger those remarks, even though the code ends up working exactly the same.

PR #66, opened 2016-04-21
Remove IDNA conversion of local-part

BreyndotEchse:
See:
zendframework/zend-mail/issues/49

### DIFF
--- a/src/EmailAddress.php
+++ b/src/EmailAddress.php
@@ -340,7 +340,7 @@ class EmailAddress extends AbstractValidator
         // atext: ALPHA / DIGIT / and "!", "#", "$", "%", "&", "'", "*",
         //        "+", "-", "/", "=", "?", "^", "_", "`", "{", "|", "}", "~"
         $atext = 'a-zA-Z0-9\x21\x23\x24\x25\x26\x27\x2a\x2b\x2d\x2f\x3d\x3f\x5e\x5f\x60\x7b\x7c\x7d\x7e';
-        if (preg_match('/^[' . $atext . ']+(\x2e+[' . $atext . ']+)*$/', $this->idnToAscii($this->localPart))) {
+        if (preg_match('/^[' . $atext . ']+(\x2e+[' . $atext . ']+)*$/', $this->localPart)) {
             $result = true;
         } else {
             // Try quoted string format (RFC 5321 Chapter 4.1.2)
--- a/src/EmailAddress.php
+++ b/src/EmailAddress.php
@@ -473,7 +473,7 @@ class EmailAddress extends AbstractValidator
         }
 
         $this->localPart = $matches[1];
-        $this->hostname  = $matches[2];
+        $this->hostname  = $this->idnToUtf8($matches[2]);
 
         return true;
     }
@@ -497,7 +497,7 @@ class EmailAddress extends AbstractValidator
         }
 
         $length  = true;
-        $this->setValue($this->idnToUtf8($value));
+        $this->setValue($value);
 
         // Split email address up and disallow '..'
         if (!$this->splitEmailParts($this->getValue())) {
--- a/src/EmailAddress.php
+++ b/src/EmailAddress.php
@@ -379,7 +379,7 @@ class EmailAddress extends AbstractValidator
     {
         $mxHosts = [];
         $weight  = [];
-        $result = getmxrr($this->idnToAscii($this->hostname), $mxHosts, $weight);
+        $result = getmxrr($this->hostname, $mxHosts, $weight);
         if (!empty($mxHosts) && !empty($weight)) {
             $this->mxRecord = array_combine($mxHosts, $weight) ?: [];
         } else {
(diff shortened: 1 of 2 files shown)


test/EmailAddressTest.php
```diff
@@ -741,8 +741,8 @@ public function testUseMxCheckBasicValid()
         ];
 
         if (extension_loaded('intl')) {
-            $emailAddresses[] = 'иван@письмо.рф';
```
Maks3w: Why modify this?

BreyndotEchse: The current validator only accepts RFC 5321 email addresses (atext = ALPHA / DIGIT / "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "/" / "=" / "?" / "^" / "_" / "`" / "{" / "|" / "}" / "~"). Because of the wrong use of IDNA for local parts, validate() returned false positives for local parts with unicode characters. When we implement [RFC 6532 #3.2](https://tools.ietf.org/html/rfc6532#section-3.2) the user should be able to choose if he wants to validate against RFC 5321 or RFC 6532 because SMTPUTF8 is not yet widely supported and just an extension

Maks3w: So this is a breaking change, as you said an option should be implemented for choose what rules should be used

BreyndotEchse: Yep. But this is part of another PR to make this validator RFC 6532 compatible (coming soon).

src/EmailAddress.php
```diff
@@ -473,7 +473,7 @@ protected function splitEmailParts($value)
         }
 
         $this->localPart = $matches[1];
-        $this->hostname  = $matches[2];
+        $this->hostname  = $this->idnToUtf8($matches[2]);
```
Maks3w: hostname should host the ascii representation as is more portable for any other use of the value.

For example the following condition `(strlen($this->hostname) > 255)` is now bugged as strlen does not allow multibyte characters (UTF-8)

test/EmailAddressTest.php
Maks3w: Please place those values which you consider invalid in the appropiate test (testUseMxRecordsBasicInvalid)

test/EmailAddressTest.php
```diff
@@ -741,8 +741,8 @@ public function testUseMxCheckBasicValid()
...
-            $emailAddresses[] = 'xn--@-7sbfxdyelgv5j.xn--p1ai';
+            $emailAddresses[] = 'test@письмо.рф';
+            $emailAddresses[] = 'test@xn--h1aigbl0e.xn--p1ai';
```
Maks3w: Why the domain part change here? You say the issue is in the local part, so why change the domain part?

BreyndotEchse: Because
```
var_dump(
    idn_to_ascii('иван@письмо.рф'),
    'test@' . idn_to_ascii('письмо.рф')
);
```
prints
```
string 'xn--@-7sbfxdyelgv5j.xn--p1ai' (length=28)

string 'test@xn--h1aigbl0e.xn--p1ai' (length=27)
```

src/EmailAddress.php
BreyndotEchse: Ouch... Maybe I should fix Hostname validator first...?

test/EmailAddressTest.php
BreyndotEchse: Added some tests for valid/invalid emails.

BreyndotEchse:
Depends on zendframework/zend-validator/pull/67. Merge that one first.

test/EmailAddressTest.php
```diff
@@ -232,6 +232,13 @@ public function validEmailAddresses()
             'bob@verylongdomainsupercalifragilisticexpialidociousspoonfulofsugar.com' => ['bob@verylongdomainsupercalifragilisticexpialidociousspoonfulofsugar.com'],
             "B.O'Callaghan@domain.com"                                                => ["B.O'Callaghan@domain.com"],
         ];
+        
+        if (extension_loaded('intl')) {
+            $return['bob@موقع.إختبار']             = ['bob@موقع.إختبار'];
+            $return['bob@xn--4gbrim.xn--kgbechtv'] = ['bob@xn--4gbrim.xn--kgbechtv'];
```
weierophinney: I've merged #67 at this point, and just attempted to merge this PR. When running tests, both of these specific cases fail, each with:

```text
not a valid hostname for the email address
The input appears to be a DNS hostname but cannot match TLD against known list
The input appears to be a local network name but local network names are not allowed
Failed asserting that false is true.
```

I'm not sure if that's _expected_ at this point, and you have more work to do, or if there's another, more subtle issue.

BreyndotEchse:
@weierophinney I used a [IANA test TLD](http://www.iana.org/domains/reserved) in this test, but Hostname validator does not accept any of these. Could have tested it by myself first... sorry :/

PR #67, opened 2016-04-26
Validate Punycoded TLDs

BreyndotEchse:
Method ``isValid`` must return true for IDN top level domains.

### DIFF
--- a/src/Hostname.php
+++ b/src/Hostname.php
@@ -1772,8 +1772,16 @@ class Hostname extends AbstractValidator
                     // id-prefix: alpha / digit
                     // ldh: alpha / digit / dash
 
-                    // Match TLD against known list
                     $this->tld = strtoupper($matches[1]);
+                    // Decode Punycode TLD to IDN
+                    if (strpos($this->tld, 'xn--') === 0) {
+                        $this->tld = $this->decodePunycode(substr($this->tld, 4));
+                        if ($this->tld === false) {
+                            return false;
+                        }
+                    }
+
+                    // Match TLD against known list
                     if ($this->getTldCheck()) {
                         if (!in_array(strtolower($this->tld), $this->validTlds)
                             && !in_array($this->tld, $this->validTlds)) {
--- a/src/Hostname.php
+++ b/src/Hostname.php
@@ -1772,13 +1772,15 @@ class Hostname extends AbstractValidator
                     // id-prefix: alpha / digit
                     // ldh: alpha / digit / dash
 
-                    $this->tld = strtoupper($matches[1]);
+                    $this->tld = $matches[1];
                     // Decode Punycode TLD to IDN
                     if (strpos($this->tld, 'xn--') === 0) {
                         $this->tld = $this->decodePunycode(substr($this->tld, 4));
                         if ($this->tld === false) {
                             return false;
                         }
+                    } else {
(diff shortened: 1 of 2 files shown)


test/HostnameTest.php
```diff
@@ -506,6 +506,25 @@ public function testAdditionalUTF8TLDs()
         }
     }
 
+    public function testAdditionalPunycodedTLDs()
+    {
+        $validator = new Hostname(Hostname::ALLOW_ALL);
+        // Check UTF-8 TLD matching
+        $valuesExpected = [
```
Maks3w: Please refactor using data providers.

Also split the test in two tests, one when is valid and another one when is not

Maks3w: When refactor please add a description for each case using the key of the array

`['invalid foo' => 'xn--3-owe4au9mpa.xn--xkc2al3hye2a']`

src/Hostname.php
```diff
@@ -1772,8 +1772,16 @@ public function isValid($value)
                     // id-prefix: alpha / digit
                     // ldh: alpha / digit / dash
 
-                    // Match TLD against known list
                     $this->tld = strtoupper($matches[1]);
+                    // Decode Punycode TLD to IDN
+                    if (strpos($this->tld, 'xn--') === 0) {
+                        $this->tld = $this->decodePunycode(substr($this->tld, 4));
```
Maks3w: Setting multibyte values in the property breaks non mutlibyte methods `if (!in_array(strtolower($this->tld), $this->validTlds)`

Maks3w: TLD attribute must be ASCII for keep the backward compatibility

test/HostnameTest.php
BreyndotEchse: Only for ``testAdditionalPunycodedTLDs`` or also for ``testAdditionalUTF8TLDs`` (merged)?

Example dataProvider (valid hostnames):
```
public function validTLDHostnames()
{
    // @codingStandardsIgnoreStart
    return [
        'test123.онлайн' => 'test123.онлайн',
        'test123.онлайн' => 'test123.xn--80asehdb',
        'тест.рф'        => 'тест.рф',
        'тест.рф'        => 'xn--e1aybc.xn--p1ai',
        'туршилтын.мон'  => 'туршилтын.мон',
        'туршилтын.мон'  => 'xn--h1aggjjdd5b4a.xn--l1acc',
    ];
    // @codingStandardsIgnoreEnd
}
```

Maks3w: I'll appreciate if you can do it testAdditionalUTF8TLDs as well.

What I miss with the descriptions I don't quickly understand why that case is significative.

```
return [
        'UTF8 TLD' => 'test123.онлайн',
        'Punyencoded TLD' => 'test123.xn--80asehdb',
...
    ];
```

src/Hostname.php
BreyndotEchse: But the current behaviour is to keep the UTF-8 TLD. For ``$hostnameValidator->isValid('тест.рф')`` the property ``$tld`` is ``рф``. Tested with latest master.

``Hostname::$validIdns`` uses UTF-8 TLDs as keys and ``Hostname::$validTlds`` holds the UTF-8 TLDs.

PR #78, opened 2016-05-11
StringLength: Added length to message variables

FabianKoestring:
Added the ability to use `%length%` as an message variable.

I think it is better to show the user how many chars he already typed if a `min` or `max` value is set. This is optional but the ability to show the value (length) aims for a better user experience. 

**For example:**
`The input is less than %min% characters long. You already typed %length% characters.` gets converted to `The input is less than 100 characters long. You already typed 99 characters.`.


vaclavvanik:
You should add test

weierophinney:
@FabianKoestring Please ping me if/when you have added tests to validate the new behavior, and I'll schedule it for a release.

Thanks!

FabianKoestring:
@weierophinney i added a test case. Is it ok or should i add something else?

samsonasik:
ping @weierophinney

PR #79, opened 2016-05-11
[Docs] Switch from Bookdown to MkDocs configuration

froschdesign:


weierophinney:
Awesome; used this as the starting point, and have now imported the various rST sources and converted them to markdown. Editing as I write this.

PR #81, opened 2016-05-16
register Uuid validator into ValidatorPluginManager

samsonasik:
in favor of #58


lowtower:
@weierophinney: can this be merged?

Issue #87, opened 2016-06-14
NotEmpty __construct creates bitmap incorrectly


AndrolGenhald:
When constructing a NotEmpty validator with an array of types, the constructor adds the bitmaps instead of bit-oring them like calculateTypeValue. It's only an issue if overlapping types such as 'php' and 'boolean' are specified, but I imagine this would be a pain to debug if anyone ever runs into it.

As shown in the documentation:

```
$validator = new Zend\Validator\NotEmpty(array(
    'boolean',
    'php'
));
```

Using the constants doesn't even work, it's ignored:

```
$validator = new Zend\Validator\NotEmpty(array(
    Zend\Validator\NotEmpty::BOOLEAN,
    Zend\Validator\NotEmpty::PHP
));
```

PR #88, opened 2016-06-14
Bugfix/not empty constructor creates bad bitmap

AndrolGenhald:
Fix issue #87

DateStepTest->testMoscowWinterTime started failing when I was working on this, but it fails on master now too.

### DIFF
--- a/src/NotEmpty.php
+++ b/src/NotEmpty.php
@@ -100,19 +100,8 @@ class NotEmpty extends AbstractValidator
         }
 
         if (is_array($options)) {
-            if (!array_key_exists('type', $options)) {
-                $detected = 0;
-                $found    = false;
-                foreach ($options as $option) {
-                    if (in_array($option, $this->constants, true)) {
-                        $found = true;
-                        $detected |= array_search($option, $this->constants);
-                    }
-                }
-
-                if ($found) {
-                    $options['type'] = $detected;
-                }
+            if (($type = $this->calculateTypeValue($options)) != 0) {
+                $options['type'] = $type;
             }
         }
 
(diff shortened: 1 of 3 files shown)


Maks3w:
The proposed changes breaks current tests. Could you review it?

test/NotEmptyTest.php
```diff
@@ -27,6 +27,33 @@ public function setUp()
         $this->validator = new NotEmpty();
     }
 
+    public function testConstructorWithTypeArray()
```
Maks3w: Please refactor this as a data provider.

test/NotEmptyTest.php
```diff
@@ -27,6 +27,33 @@ public function setUp()
...
+    {
+        $validator = new NotEmpty([
+            'php',
```
Maks3w: Could you replace php|boolean example with Integer|Zero I think those values represent better the bitmask behavior.

AndrolGenhald: Not sure what you mean here, the integer and zero bitmasks don't overlap, so they don't cause the issue that I'm fixing.

Maks3w:
The coverage drop is normal, less lines per file leads to a higher percent per (uncovered) line.

src/NotEmpty.php
```diff
@@ -100,19 +100,8 @@ public function __construct($options = null)
...
-                        $found = true;
-                        $detected += array_search($option, $this->constants);
-                    }
-                }
-
-                if ($found) {
-                    $options['type'] = $detected;
-                }
+            if (($type = $this->calculateTypeValue($options)) != 0) {
+                $options['type'] = $type;
```
Maks3w: It's better create a new $options array here for to clean up from those old values are now represented by $type.

Maks3w:
I appreciate the constructor needs more improvements. `calculateTypeValue` is called at least 3 times.

1. $this->setType($this->defaultType);
2. $this->calculateTypeValue($options));
3. parent::__construct($options); // if $options has the option key `type`

Could you take a look to this?

src/NotEmpty.php
AndrolGenhald: This would break options for AbstractValidator. I checked and it doesn't actually cause testTypeAutoDetectionHasNoSideEffect to fail, so maybe another test should be added that uses an array that has both types and AbstractValidator options? Either that or document that if the array has any non-type values it will just be treated as a normal options array.
It looks like the current constructor allows type strings to be combined with other options.

AndrolGenhald:
`calculateTypeValue` should be called at most twice now, where the second call will only check 2 conditionals before returning. I don't think it's possible to guarantee that it'll only be called once, as there has to be some way to tell if the options array directly contains types.

Issue #90, opened 2016-06-20
Zend\Validator\File\ExcludeMimeType is missing the message templates


Jakovitz:
There are no error message templates in Zend\Validator\File\ExcludeMimeType.

samsonasik:
it extends `Zend\Validator\File\MimeType`:

```php
class ExcludeMimeType extends MimeType
```
so, no need to re-define the $messageTemplates

webimpress:
Unfortunately not, because in `Zend\Validator\File\MimeType` we have:

```
    protected $messageTemplates = [
        self::FALSE_TYPE   => "File has an incorrect mimetype of '%type%'",
        self::NOT_DETECTED => "The mimetype could not be detected from the file",
        self::NOT_READABLE => "File is not readable or does not exist",
    ];
```

so it is always using constants from `MimeType`.

We have the same constant names in `ExcludeMimeType`, but different constant values and we are using `self` in definition of the message template array. So @Jakovitz has right. It's a bug.

We can't use `static` instead of `self` in message template array definition (because of `"static::" is not allowed in compile-time constants`) so probably array have to be redefined in `ExcludeMimeType` ...

samsonasik:
this can be closed then

samsonasik:
can be closed in favor of https://github.com/zendframework/zend-validator/issues/92

PR #92, opened 2016-06-20
Fix - missing `messageTemplates` in `ExcludeMimeType` validator

webimpress:
Fix for bug reported in #90

### DIFF
--- a/src/File/ExcludeMimeType.php
+++ b/src/File/ExcludeMimeType.php
@@ -22,6 +22,15 @@ class ExcludeMimeType extends MimeType
     const NOT_READABLE = 'fileExcludeMimeTypeNotReadable';
 
     /**
+     * @var array Error message templates
+     */
+    protected $messageTemplates = [
+        self::FALSE_TYPE   => "File has an incorrect mimetype of '%type%'",
+        self::NOT_DETECTED => "The mimetype could not be detected from the file",
+        self::NOT_READABLE => "File is not readable or does not exist",
+    ];
+
+    /**
      * Returns true if the mimetype of the file does not matche the given ones. Also parts
      * of mimetypes can be checked. If you give for example "image" all image
      * mime types will not be accepted like "image/gif", "image/jpeg" and so on.
--- a/test/File/ExcludeMimeTypeTest.php
+++ b/test/File/ExcludeMimeTypeTest.php
@@ -51,6 +51,10 @@ class ExcludeMimeTypeTest extends \PHPUnit_Framework_TestCase
         $validator = new ExcludeMimeType($options);
         $validator->enableHeaderCheck();
         $this->assertEquals($expected, $validator->isValid($isValidParam));
+        if (!$expected) {
+            $this->assertArrayHasKey($validator::FALSE_TYPE, $validator->getMessages());
+            $this->assertNotEmpty($validator->getMessages()[$validator::FALSE_TYPE]);
+        }
     }
 
     /**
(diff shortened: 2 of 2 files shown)


test/File/ExcludeMimeTypeTest.php
```diff
@@ -51,6 +51,10 @@ public function testBasic($options, $isValidParam, $expected)
         $validator = new ExcludeMimeType($options);
         $validator->enableHeaderCheck();
         $this->assertEquals($expected, $validator->isValid($isValidParam));
+        if (!$expected) {
```
Maks3w: Please instead add the expected getMessages return value on the data provider.

test/File/ExcludeMimeTypeTest.php
```diff
@@ -44,13 +48,18 @@ public function basicBehaviorDataProvider()
...
      * @dataProvider basicBehaviorDataProvider
-     * @return void
+     *
+     * @param string|array $options
+     * @param array $isValidParam
+     * @param bool $expected
+     * @param array $messages
      */
-    public function testBasic($options, $isValidParam, $expected)
+    public function testBasic($options, array $isValidParam, $expected, array $messages = [])
```
Maks3w: It's better to have all possible values in the data provider for get the whole picture. Please remove the default value.

test/File/ExcludeMimeTypeTest.php
```diff
@@ -25,17 +25,21 @@ public function basicBehaviorDataProvider()
...
-            //    Options, isValid Param, Expected value
-            ['image/gif',                      $fileUpload, true],
-            ['image',                          $fileUpload, false],
-            ['test/notype',                    $fileUpload, true],
-            ['image/gif, image/jpeg',          $fileUpload, false],
+            //    Options, isValid Param, Expected value, messages
+            ['image/gif',                 $fileUpload, true],
+            ['image',                     $fileUpload, false, [ExcludeMimeType::FALSE_TYPE => "File has an incorrect mimetype of 'image/jpeg'"]],
+            ['test/notype',               $fileUpload, true],
+            ['image/gif, image/jpeg',     $fileUpload, false, [ExcludeMimeType::FALSE_TYPE => "File has an incorrect mimetype of 'image/jpeg'"]],
```
Maks3w: You can avoid repeat the same message using a local variable `$falseTypeMessage`

Maks3w: https://github.com/zendframework/zend-inputfilter/blob/master/test/InputTest.php#L688

This is an example of a readable data provider. Notice even we mask the scalar values with a variable name so you can understand the meaning inmediatelly.

src/File/ExcludeMimeType.php
```diff
@@ -22,6 +22,15 @@ class ExcludeMimeType extends MimeType
     const NOT_READABLE = 'fileExcludeMimeTypeNotReadable';
 
     /**
+     * @var array Error message templates
+     */
+    protected $messageTemplates = [
+        self::FALSE_TYPE   => "File has an incorrect mimetype of '%type%'",
```
Maks3w: I think INCORRECT_TYPE it's a better name. Why FALSE?

webimpress: Constant was defined, the same constants name are in parent class `MimeType`, I've added only messages.
We can change it, maybe it is even better, but for consistency I'd prefer to keep `FALSE_TYPE` ...
And it's not in the scope of the issue ;)

Maks3w: ok. You're right.

weierophinney:
@Maks3w Please don't forget to update the CHANGELOG when merging... (I'll do it now, so I can release this bugfix.)

Maks3w:
I was waiting for add more patches to the release

weierophinney:
No need to wait on releases anymore! :-) My rule of thumb is that if I
won't complete reviewing patches in the same session, release immediately,
and do the other patches in the next release. This is also why it's
important to do change logs on merge; it ensures nobody has to look up
what's been merged when doing a release.

PR #94, opened 2016-06-23
fixed a bug where empty $email was passed to idn_to_utf8

Exlord:
checked if $email is not empty before passing it to idn_to_utf8.
idn_to_utf8 throws a warning when value is empty

### DIFF
--- a/src/EmailAddress.php
+++ b/src/EmailAddress.php
@@ -547,6 +547,9 @@ class EmailAddress extends AbstractValidator
      */
     protected function idnToUtf8($email)
     {
+        if(strlen($email) == 0)
+            return $email;
+            
         if (extension_loaded('intl')) {
             // The documentation does not clarify what kind of failure
             // can happen in idn_to_utf8. One can assume if the source
--- a/src/EmailAddress.php
+++ b/src/EmailAddress.php
@@ -547,8 +547,9 @@ class EmailAddress extends AbstractValidator
      */
     protected function idnToUtf8($email)
     {
-        if (strlen($email) == 0)
+        if (strlen($email) == 0) {
             return $email;
+        }
             
         if (extension_loaded('intl')) {
             // The documentation does not clarify what kind of failure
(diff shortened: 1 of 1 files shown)


Exlord:
Test case already exist for empty value
https://github.com/zendframework/zend-validator/blob/master/test/EmailAddressTest.php#L259
but this only happens when `intl` extension is active.

src/EmailAddress.php
```diff
@@ -547,6 +547,9 @@ protected function idnToAscii($email)
      */
     protected function idnToUtf8($email)
     {
+        if(strlen($email) == 0)
+            return $email;
```
webimpress: PSR-2, missing parentheses and space after `if`

http://www.php-fig.org/psr/psr-2/

src/EmailAddress.php
```diff
@@ -547,6 +547,9 @@ protected function idnToAscii($email)
      */
     protected function idnToUtf8($email)
     {
+        if (strlen($email) == 0)
```
webimpress: PSR-2 missing opening and closing parentheses: http://www.php-fig.org/psr/psr-2/#5-1-if-elseif-else

webimpress:
Could you also provide test for your change?

src/EmailAddress.php
Exlord: braces where missing not parentheses.

Exlord:
@webimpress Test case already exist for empty value
https://github.com/zendframework/zend-validator/blob/master/test/EmailAddressTest.php#L259
but this only happens when intl extension is active. how can i enable intl extension when writing test case?

src/EmailAddress.php
webimpress: @Exlord sorry, you have right :) thanks for your fix

Maks3w:
1. @Exlord Please review your PR until all checks are successful.

2. INTL is enabled on CI so the bug should have been detected. So that test is not working as expected or don't cover this use case.

weierophinney:
Thanks, @Exlord

PR #105, opened 2016-07-18
Update Code128.php

FraGoTe:
Avoiding warning in the last character https://github.com/zendframework/zend-barcode/issues/28

### DIFF
--- a/src/Barcode/Code128.php
+++ b/src/Barcode/Code128.php
@@ -180,7 +180,7 @@ class Code128 extends AbstractAdapter
                     break;
             }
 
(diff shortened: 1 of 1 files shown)


Xerkus:
i will keep this open for now. May be this issue should be addressed in zend-stdlib
https://github.com/zendframework/zend-stdlib/issues/73

weierophinney:
Thanks, @FraGoTe

weierophinney:
Actually, I need to revert this.

It causes the `testCODE128()` case to enter an infinite loop.

PR #113, opened 2016-09-02
Callback validator consumes string class name via ::class resolution

vaclavvanik:
Added ability to inject string class name into callback validator constructor.

### DIFF
--- a/doc/book/validators/callback.md
+++ b/doc/book/validators/callback.md
@@ -123,6 +123,26 @@ if ($valid->isValid($input)) {
 }
 ```
 
+AS of PHP 5.5 you can use ::class resolution for given callback class.
+
+```php
+class MyClass
+{
+    public function __invoke($value)
+    {
+        // some validation
+        return true;
+    }
+}
+
+$valid = new Zend\Validator\Callback(MyClass::class);
+if ($valid->isValid($input)) {
+    // input appears to be valid
+} else {
+    // input is invalid
+}
+```
+
 ## Adding options
 
 `Zend\Validator\Callback` also allows the usage of options which are provided as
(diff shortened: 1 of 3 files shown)


doc/book/validators/callback.md
```diff
@@ -123,6 +123,26 @@ if ($valid->isValid($input)) {
...
+class MyClass
+{
+    public function __invoke($value)
+    {
+        // some validation
+        return true;
+    }
+}
+
+$valid = new Zend\Validator\Callback(MyClass::class);
```
Ocramius: This needs the method to be static...

src/Callback.php
```diff
@@ -74,6 +74,10 @@ public function getCallback()
      */
     public function setCallback($callback)
     {
+        if (is_string($callback) && class_exists($callback)) {
+            $callback = new $callback();
```
Ocramius: This sort of magic makes it unclear when DI can be used or not

doc/book/validators/callback.md
Ocramius: Nvm, was confused, then read the code for it. This breaks the principle of least possible astonishment for me.

vaclavvanik:
Im not happy that `setCallback` brings more magic. However anonymous functions cannot be serialized (and pulled from merged application config), this PR brings possibility to bypass serialization restriction. I hope that reason for this PR is more obvious.

vaclavvanik:
ping @weierophinney

vaclavvanik:
Pinging @Ocramius and @weierophinney for review.

Ocramius:
As mentioned above, I don't like this auto-instantiating approach. Most callbacks require some sort of DB connection, lookup system, etc, so they will unlikely fit this particular validator.

Generally 👎 from me, sorry.

Xerkus:
Anonymous functions can be injected via `Module::getValidatorConfig()` in zend-modulemanager based projects,  besides your specific use case would be better with static `['Class', 'method']` callable anyway.

:-1: from me as well, sorry

PR #132, opened 2016-10-20
Clean benign errors

codisart:


Xerkus:
That method can be used/overriden by extending code so changing its signature will be a BC break.

It would be better to default `$hostname` to true and simplify following check to:
```
if ($local && $hostname && $length) {
    return true;
}
```

codisart:
Thank you @Xerkus for you review. I did changes to my code.

weierophinney:
Thanks, @codisart

PR #136, opened 2016-11-15
Fix for upcoming new PHP versions

remicollet:


remicollet:
FYI, discovered in Fedora QA (which now uses PHP 7.1)
https://apps.fedoraproject.org/koschei/package/php-zendframework-zend-validator?collection=f26
(goes "red" when 7.1.0RC6 appear, goes "green" with this patch applied)

PR #138, opened 2016-11-25
Allow messages to also be translated by $messageKey

delolmo:
Currently, message keys are never used to translate template messages. Even though there are different opinions on whether the message keys and/or the message templates are immutable, there seems to be no reason as to why using both approaches to translate messages is incompatible. The proposed change 
defaults to the current behavior. However, if the returned string is identical to the one provided to the translator, a new attempt is made using the message key.


weierophinney:
@delolmo Could you please add tests covering this new functionality? Otherwise it's possible somebody would revert it later, and not realize it was intended, as the tests would still pass.

Thanks!

weierophinney:
There's a very good reason why message keys are not used: we were having conflicts between the various validators, as they were often defining the same message key, but mapping it to an entirely different message. We determined it made more sense to have a single source of truth - the message template.

As such, closing. Thanks for your efforts, @delolmo .

PR #140, opened 2016-12-10
Fix - wrong `messageTemplate` in `GreaterThan` validator

fermius:
_messageTemplate_ would be

"The input is not greater or equal than '%min%'"

instead of

"The input is not greater than or equal to '%min%'"

accordingly to [ES translation](https://github.com/zendframework/zend-i18n-resources/blob/master/languages/es/Zend_Validate.php#L205) and others


Xerkus:
`The input is not greater than or equal to '%min%'` would be the correct one. This issue should be opened on i18n resources repository instead

Issue #147, opened 2017-02-18
Zend\Math is required by Zend\Validator\Csrf but packae is not required


heiglandreas:
```Zend\Validator\Csrf``` requires the packages ```zend-session``` and ```zend-math``` but in composer.json they are only set as dev-requirements. Even though they are listed as suggestions, the package in itself is not usable without these packages. so they should be included in the ```require```-section.

weierophinney:
All validators are opt -in, and the majority have no dependencies outside those in the "require" section. For this reason, we list any validator-specific deps as suggestions, but do not require them. 

Eventually, we may split these into separate packages so we can make those deps requirements, but for now, they are suggestions.

heiglandreas:
Then the suggestions should at least include which validators can only be used when the suggested package is added…

weierophinney:
Would you be willing to do a PR for that, or at least start one?

heiglandreas:
Already at it 😉

PR #148, opened 2017-02-18
Adds more info to suggestions

heiglandreas:
This commit adds more infos why separate packages are suggested. This is especially of interest as different Validators are only usable with additional packages like the Csrf- and the (No)RowExists-validator

This addresses #147

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,13 +32,13 @@
         "phpunit/PHPUnit": "^4.0"
     },
     "suggest": {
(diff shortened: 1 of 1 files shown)


Ocramius:
@heiglandreas merged, thanks!

`master`: zendframework/zend-validator@6c3f4c03e54ffd8426ca646f363de8b43b83b443
`develop`: zendframework/zend-validator@2c659f743adf9078477b880110526381294317af

PR #150, opened 2017-02-24
Changes done for adding support for "ø" for DK idn

damanikrish:
When ever I try to create `Uri` object for url string "http://www.fitnessø.dk/", it throws an exception with message "Host "www.fitnessø.dk" is not valid or is not accepted by...". And when debugged the source, I came to know that this error is because it is not validated by HostName validator in `Uri::isValidDnsHostname()`.


akrabat:
Travis failure is due to https://github.com/zendframework/zend-servicemanager/issues/182, not this PR.

akrabat:
Are there any other characters that should be allowed in a DK url?

damanikrish:
@akrabat , I am not sure for the other characters, I just came across the use case for this character and hence created this PR for the same.

Thanks,
Krish Damani

damanikrish:
@akrabat , I have added two more characters and updated the PR,
Reference for allowed characters in dk domain names https://www.dk-hostmaster.dk/sites/default/files/Characters%20allowed%20in.pdf

Thanks,
Krish Damani

PR #154, opened 2017-03-20
Fixes #153

### LINKED ISSUE #153: Validating VISA credit cards fails if card number length is not equal to 16.
According to this [Wikipedia Page](https://en.wikipedia.org/wiki/Payment_card_number) length of a Visa card number can be 13, 16 and 19. 

I recognized this issue when I used [fzaninotto/faker](https://github.com/fzaninotto/Faker) in my PhpUnit test script, it failed for the times when the generated number is like 4909117994192. It uses templates for the Visa card number generation listed below:
```php
[
    'Visa' => array(
            "4539########",
            "4539###########",
            "4556########",
            "4556###########",
            "4916########",
            "4916###########",
            "4532########",
            "4532###########",
            "4929########",
            "4929###########",
            "40240071####",
            "40240071#######",
            "4485########",
            "4485###########",
            "4716########",
            "4716###########",
            "4###########",
            "4##############"
    )
];
```

It is reported that length of most Visa cards' number is 16. If this is why only 16 selected for Visa cards, that's ok. If not I would like to send PR for this issue.

mkorkmaz:
Credit card number lengths are extended for VISA and DISCOVER. This PR also fixes #153. 

Source: [https://en.wikipedia.org/wiki/Payment_card_number](https://en.wikipedia.org/wiki/Payment_card_number)

### DIFF
--- a/src/CreditCard.php
+++ b/src/CreditCard.php
@@ -83,14 +83,14 @@ class CreditCard extends AbstractValidator
         self::AMERICAN_EXPRESS => [15],
         self::DINERS_CLUB      => [14],
         self::DINERS_CLUB_US   => [16],
-        self::DISCOVER         => [16],
+        self::DISCOVER         => [16, 19],
         self::JCB              => [15, 16],
         self::LASER            => [16, 17, 18, 19],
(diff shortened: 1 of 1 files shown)


weierophinney:
Thanks, @mkorkmaz

PR #156, opened 2017-03-25
add IsCountable validator, to enforce type and count elements.

abacaphiliac:
hi Team,

i created something like this in a private project (to enforce a countable has at least one element) but thought a more configurable version might be useful to others. what do you think?


abacaphiliac:
closing in favor of https://github.com/zendframework/zend-validator/pull/157

PR #157, opened 2017-03-25
add IsCountable validator, to enforce type and count elements.

abacaphiliac:
hi Team,

i created something like this in a private project (to enforce a countable has at least one element) but thought a more configurable version might be useful to others. what do you think?


src/IsCountable.php
```diff
@@ -0,0 +1,150 @@
...
+     * Returns true if and only if $value is countable (and the count validates against optional values).
+     *
+     * @param  string $value
+     * @return bool
+     * @throws Exception\RuntimeException
+     */
+    public function isValid($value)
+    {
+        if (! (is_array($value) || $value instanceof \Countable)) {
+            throw new RuntimeException($this->messageTemplates[self::NOT_COUNTABLE]);
```
weierophinney: Invalid input is not an exceptional case, and should instead be reported as a validation error. Use the `$this->error` construct and return false in this particular case.

src/IsCountable.php
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        $count = count($value);
+
+        if (is_numeric($this->options['count'])) {
```
weierophinney: I'd recommend using your `get*()` methods for these, instead of direct access to the `$options` property. The rationale is that this implementation detail may change for version 3 (see the [v3 RFC](https://discourse.zendframework.com/t/rfc-zend-validator-v3/208/)), and refactoring would require finding these references throughout the class.

weierophinney: Also, it looks like the validator will have different behavior based on whether `count`, `min`, or `max` are set; in fact, it appears that these should be _mutually exclusive_. As such, I wonder if the constructor and/or `setOptions()` should have some logic to prevent cases where one is already set/discovered, and an attempt is made to set a different option?

src/IsCountable.php
```diff
@@ -0,0 +1,150 @@
...
+        return $this->options['count'];
+    }
+
+    /**
+     * Sets the count option
+     *
+     * @param  int $count
+     * @return self Provides a fluent interface
+     */
+    public function setCount($count)
```
weierophinney: Should the class offer setters? Once one of either `count`, `min`, or `max` is set, I'd argue that the class shouldn't allow setting one of the other options, as that indicates a change in behavior, and/or an unexpected behavior (e.g., if `count` was previously set, but `setMin()` or `setMax()` is called later, it will still only honor the `count` setting).

abacaphiliac:
thank you for the feedback. i'll get this fixed up tonight.

test/IsCountableTest.php
```diff
@@ -0,0 +1,88 @@
...
+    public function testInvalidNotEquals()
+    {
+        $this->sut->setCount(2);
+
+        self::assertFalse($this->sut->isValid(['Foo']));
+        self::assertCount(1, $this->sut->getMessages());
+    }
+
+    /**
+     * @expectedException \Zend\Validator\Exception\RuntimeException
```
froschdesign: Please the `expectException()` method, this is the preferred way.

test/IsCountableTest.php
```diff
@@ -0,0 +1,88 @@
...
+    {
+        $this->sut = new IsCountable();
+    }
+
+    public function testArrayIsValid()
+    {
+        $this->sut->setMin(1);
+        $this->sut->setMax(10);
+
+        self::assertTrue($this->sut->isValid(['Foo']), json_encode($this->sut->getMessages()));
```
froschdesign: Please use `$this->` instead of `self::`, this is the preferred way.

webimpress: @Ocramius thinks something else 😜

abacaphiliac:
thanks everyone for the review. i think i took care of everything asked. @weierophinney i chose to implement the mutex logic in setOptions because it is invoked from the constructor. my test case covers behavior of the constructor and setOptions.

weierophinney:
@abacaphiliac I've added some commits that do the following:

- Make having `count` and one or both of `min` or `max` mutually exclusive; setting a `count` when either `min` or `max` are defined, or setting `min` or `max` when `count` is defined will raise an exception now. This prevents problems issues when you merge configuration; if you expect one set of behavior, you should not have a different behavior!
- Adds documentation.

I'm expecting the only test failure to be due to the hostname TLD list being out-of-date (again! those used to update more on an order of weeks or months, and now it's practically daily!); if that's the case, I can merge this.

weierophinney:
Thanks for this great feature, @abracaphiliac\! I have merged it to develop for an upcoming 2.10.0 release (likely today or tomorrow).

abacaphiliac:
my pleasure. thank you for the assist : )

PR #160, opened 2017-04-05
Support for internationalized email address local part

Xerkus:
[RFC 6532](https://tools.ietf.org/html/rfc6532#section-3.1) extends atext used in local part with non-ascii unicode support

I have no idea what am i doing here. It seems to be working, but it would be great if someone understanding  unicode could verify.


src/EmailAddress.php
```diff
@@ -342,6 +344,8 @@ protected function validateLocalPart()
         $atext = 'a-zA-Z0-9\x21\x23\x24\x25\x26\x27\x2a\x2b\x2d\x2f\x3d\x3f\x5e\x5f\x60\x7b\x7c\x7d\x7e';
         if (preg_match('/^[' . $atext . ']+(\x2e+[' . $atext . ']+)*$/', $this->localPart)) {
             $result = true;
+        } elseif ($this->validateInternationalizedLocalPart()) {
```
weierophinney: To make this a bit more robust, please pass `$this->localPart` as an argument to `validateInternationalizedLocalPart()`. When we make the validators stateless for v3, this will then be one less method we need to change. :smile:

weierophinney:
Thanks, @Xerkus; merged to develop for release with 2.10.0.

Xerkus:
ugh, what is going on here with the commits and comments?

PR #166, opened 2017-04-25
Ref #7629

philippe-jung:
Closes opened file with finfo to avoid segfault on specific PHP versions

See https://github.com/zendframework/zendframework/issues/7629 for reproduction details


weierophinney:
Thanks, @philippe-jung

Issue #165, opened 2017-04-13
MimeType Validator Not Working


open-it:
To check if a file is readable the MimeType Validator does a stream_resolve_include_path($file) (line 366), as my image has been uploaded to my /tmp directory (which isn't in my include_path), I can never validate my image.

froschdesign:
Duplicates #77 (See also this PR: #169)

Issue #77, opened 2016-05-11
[question] Why do the File\Extension validator validate the presence of a file?


larsnystrom:
Is it really the job of the `File\Extension` validator to check that the file exists? Especially given that there already is a `File\Exists` validator.

I'm trying to write some tests for a class which inherits `File\Extension`, but the fact that the validator checks that the file exists makes it harder than I think it should be.

My problem is with these lines:

```
    // Is file readable ?
    if (empty($file) || false === stream_resolve_include_path($file)) {
        $this->error(self::NOT_FOUND);
        return false;
    }
```

A related question is why `File\Extension` uses `stream_resolve_include_path` instead of `is_readable` to check that the file exists. I can't think of a use case where you'd want to look in the include paths. To me it looks borderline security related, even though I can't think of a concrete threat right now.

PR #169, opened 2017-05-23
Use is_readable() to check if file is readable

larsnystrom:
This solves #165 and #77. 

Sorry about mixing the formatting and the fix.


weierophinney:
Scheduling for 2.10.0, as it's a slight (though justified) behavior change.

weierophinney:
Thanks, @larsnystrom. I have merged to develop for release with 2.10.0, which should occur this week or next.

PR #173, opened 2017-06-15
Corrected Date validator format docs

### LINKED ISSUE #158: Misleading documentation on accepted formats by Date Validator
In the documentation it is mention that the `format` option accepts any format allowed by the PHP `date()` function. This seems not to be the case since it seems to be using the method `DateTime::createFromFormat` in order to validate the date, and this only accepts a subset of the formats that `date()` does.

For example a perfectly valid format such as `N` to get the day of the week as an integer from 1 to 7 would not be valid for the Date Validator.

nathanjosiah:
Fixes #158


weierophinney:
Thanks, @nathanjosiah

PR #174, opened 2017-06-19
Fix Between validator behavior

### LINKED ISSUE #167: Between Validator outputs true for characters
Hi,

I am using version "2.9.0 - 2017-03-17" of zendframework/zend-validator.
If I input string "asdasd" for field "price" validator returns true. Expecting: false.

String "asdasd" is not even a number. Therefore it cannot lie between 0 and 99999999.
<pre>

public function init() {
     $this->add(array(
            'type' => 'text',
            'name' => 'price',
            'options' => array(
                'label' => $this->translator->translate('label.price', 'Form'),
                'label_attributes' => array(
                    'class' => 'col-sm-3',
                ),
                'column-size' => 'sm-5',
            ),
            'attributes' => array(
                'id' => 'price',
            ),
        ));
}

public function getInputFilterSpecification()
    {
        return array(
            'price' => array(
                'required' => true,
                'validators' => array(
                    array(
                        'name' => 'Between',
                        'options' => array(
                            'min' => 0,
                            'max' => 99999999,
                            'inclusive' => true
                        ),
                    ),
                ),
            ),
        );
    }
</pre>

codisart:
fixes #167 

return false when validating a string with a numeric range
return false when validating a numeric value with a string range


src/Between.php
```diff
@@ -17,6 +17,8 @@ class Between extends AbstractValidator
     const NOT_BETWEEN        = 'notBetween';
     const NOT_BETWEEN_STRICT = 'notBetweenStrict';
 
+    private $nuumeric;
```
samsonasik: typo, `nuumeric` should be `numeric`, surplus `u` char

codisart: Thank you for you review @samsonasik !

src/Between.php
```diff
@@ -17,6 +17,8 @@ class Between extends AbstractValidator
     const NOT_BETWEEN        = 'notBetween';
     const NOT_BETWEEN_STRICT = 'notBetweenStrict';
 
+    private $numeric;
```
froschdesign: Please add always a DocBlock with a short description and data type.

codisart: Thank you for your review @froschdesign

src/Between.php
```diff
@@ -162,6 +179,13 @@ public function setInclusive($inclusive)
      */
     public function isValid($value)
     {
+        if ($this->numeric && ! is_numeric($value)) {
+            return false;
+        }
+        if (! $this->numeric && ! is_string($value)) {
+            return false;
+        }
```
weierophinney: Each of these should be calling `$this->error()` in order to set a validation failure message; otherwise, there's no way to understand _why_ validation failed. Take a look at how the other cases that result in failure work, if you are unsure how to proceed.

test/BetweenTest.php
```diff
@@ -18,37 +18,76 @@
...
 {
+    public function providerBasic()
+    {
+        return [
+            [
+                'min' => 1,
+                'max' => 100,
+                'inclusive' => true,
+                'expected' => true,
+                'values' => [1, 10, 100],
```
weierophinney: If you're going to break this into a data provider, do a separate set per value as well.

Additionally, give each set a key with a descriptive name:

```php
'inclusive-valid-floor' => [
    'min' => 1,
    'max' => 100,
    'inclusive' => true,
    'expected' => true,
    'value' => 1,
],
'inclusive-valid-in-range' => [
    'min' => 1,
    'max' => 100,
    'inclusive' => true,
    'expected' => true,
    'value' => 1,
],
'inclusive-valid-ceiling' => [
    'min' => 1,
    'max' => 100,
    'inclusive' => true,
    'expected' => true,
    'value' => 100,
],
```

This makes it easy to identify which case to examine when a failure occurs, as PHPUnit will report the key; you can then search for that key to identify each set of data provided.

src/Between.php
weierophinney: Also, once you do make this change, add tests for each condition to validate that the expected failure messages are present.

weierophinney (changes requested):
This looks good, but needs a few changes before I can merge it.

weierophinney (approved):

weierophinney:
@codisart I've made the changes I requested, and pushed them to your branch so as to verify the pull request against Travis. Merging now so we can do a new 2.9 series release.

Thanks!

PR #180, opened 2017-07-18
Fix BC break introduced by 3fc1e66

### LINKED ISSUE #170: MimeType validator doesn't work for multiple files.
Hi,

The IsImage validator doesn't work with html5 file inputs that have the multiple attribute assigned.

It validates the first file file, then gives the error : 

Warning: finfo_file(): 557 is not a valid file_info resource in vendor/zendframework/zend-validator/src/File/MimeType.php on line 387

Warning: finfo_close(): 557 is not a valid file_info resource in vendor/zendframework/zend-validator/src/File/MimeType.php on line 388

Ideally, the behaviour is that it won't pass validation unless all of the files are images?

Thanks

### LINKED ISSUE #176: Mimetype-Validation of Multiple Files broken
https://github.com/zendframework/zend-validator/commit/3fc1e6614672eb7dbba816677c5a660a953bb563

This currently Breaks Mimetype-Validation of Multiple Files because it closes the finfo-handle after the first validation. 
At the second Validation the finfo-handle is closed and does not reopen at the setMagicFile-Function.

ThomasCantonnet:
Fixes #170 and #176 by using finfo's destructor via unset:

* https://github.com/php/php-src/blob/0437aa2abf1eb494e4555d7d4dfbdbf566a59bdc/ext/fileinfo/fileinfo.c#L185
* https://github.com/php/php-src/blob/0437aa2abf1eb494e4555d7d4dfbdbf566a59bdc/ext/fileinfo/tests/finfo_close_basic.phpt

Thus avoiding the call to an invalid finfo resource while still avoiding the segfault bug on older PHP versions.

### DIFF
--- a/src/File/MimeType.php
+++ b/src/File/MimeType.php
@@ -385,7 +385,7 @@ class MimeType extends AbstractValidator
             $this->type = null;
             if (! empty($this->finfo)) {
                 $this->type = finfo_file($this->finfo, $file);
-                finfo_close($this->finfo);
+                unset($this->finfo);
             }
         }
 


ThomasCantonnet:
@weierophinney could we at the very least rollback the commit ? The MimeType validator is not usable anymore for multiple file validation.

froschdesign:
@ThomasCantonnet 
This will not fix #170, because the problem is different.

ThomasCantonnet:
@froschdesign IsImage is calling the MimeType validator and errors on the same line, did I miss something?

froschdesign:
@ThomasCantonnet 
> …did I miss something?

The problem in #170 is related to multiple files and the entire `$_FILES` array, but this is not supported.

ThomasCantonnet:
@froschdesign right this is kind of a side effect to #176 then, the issue remains even without the error i suppose?

froschdesign:
@ThomasCantonnet 
Not a side effect, because the `$_FILES` array can include multiple file values, but this is not supported. So, this is an entirely different problem.

weierophinney:
Thanks, @ThomasCantonnet

PR #186, opened 2017-07-28
add cn and biz hostname validator tests

abacaphiliac:
hi Team,

i noticed the red coverage badge on the readme and thought i'd help. ran a coverage report and saw that most of the interesting bits are already covered! almost all validators are at 90%+; great job!

i guess we're getting dinged by the large cn and biz TLD files that were probably extracted from the hostname validator. these two tests will replace that distracting red coverage badge with something a little more representative of the work you've accomplished here.


weierophinney:
Thanks, @abacaphiliac

PR #188, opened 2017-08-02
ValidatorManager as alias instead of factory identifier

boesing:
As I like the usage of `ValidatorPluginManager::class` notation, I would prefer if services like the `ValidatorManager` will be able to get fetched by using `ValidatorPluginManager::class` aswell.

Is there any reason why we wont provide such ability?


weierophinney:
Thanks, @boesing. Merged to develop for release with 2.10.0

PR #189, opened 2017-08-02
Added `ValidatorPluginManagerAwareInterface` to `ValidatorChain`

boesing:
This marks `ValidatorChain::setPluginManager` and `ValidatorChain::getPluginManager` as deprecated while proxying to `ValidatorPluginManagerAwareInterface` methods

### DIFF
--- a/src/ValidatorChain.php
+++ b/src/ValidatorChain.php
@@ -80,7 +80,7 @@ class ValidatorChain implements
      */
     public function setPluginManager(ValidatorPluginManager $plugins)
     {
-        $this->plugins = $plugins;
+        $this->setValidatorPluginManager($plugins);
         return $this;
     }
 
--- a/src/ValidatorChain.php
+++ b/src/ValidatorChain.php
@@ -15,7 +15,8 @@ use Zend\ServiceManager\ServiceManager;
 
 class ValidatorChain implements
     Countable,
-    ValidatorInterface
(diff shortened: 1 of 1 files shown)


weierophinney:
What benefit does this provide?

Starting with the v3 components, we're moving _away from_ the usage of `*AwareInterface`, as these require usage of initializers, which we have deprecated. I see this as a step backwards, particularly as the `setPluginManager()` method has been defined since initial release.

boesing:
@weierophinney thanks for that explanation. So, since `*AwareInterface` (initializer) logic is being dropped, will there be an easy method to configure setter injection? 

Just wondering, since writing a delegator for any use-case would lead to pretty much code, just in favor of avoiding aware interfaces.

Is there any "preview" or discussion I can follow to keep up to date with things which will change in upcoming versions?

weierophinney:
> Just wondering, since writing a delegator for any use-case would lead to pretty much code, just in favor of avoiding aware interfaces.

Writing a delegator is no more code than an initializer. It's a bit more configuration, but the delegator factory itself can be re-used across multiple services.

In terms of discussion, the zend-servicemanager changes were completed around 18 months ago, and we've had two additional minor releases of that library since. For ongoing discussion, we have:

- [Slack](https://zendframework-slack.herokuapp.com)
- [Forums](https://discourse.zendframework.com)

PR #194, opened 2017-08-19
Fixes #193 : use INTL_IDNA_VARIANT_UTS46 constant only when exists

### LINKED ISSUE #193: php 7.1.8 with zend-validator 2.10.0 : undefined constant INTL_IDNA_VARIANT_UTS46
in EmailAddress.php: 557

https://github.com/zendframework/zend-validator/blob/33caf7f7ab2cc3181895820b2fe17bd792f72226/src/EmailAddress.php#L556-L558

Need to check if constant available ?

samsonasik:
Fixes #193

### DIFF
--- a/bin/update_hostname_validator.php
+++ b/bin/update_hostname_validator.php
@@ -177,9 +177,12 @@ function getNewValidTlds($string)
 function getPunycodeDecoder()
 {
     if (function_exists('idn_to_utf8')) {
-        return function ($domain) {
-            return idn_to_utf8($domain, 0, INTL_IDNA_VARIANT_UTS46);
-        };
+        if (defined('INTL_IDNA_VARIANT_UTS46')) {
+            return function ($domain) {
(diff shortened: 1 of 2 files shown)


bin/update_hostname_validator.php
```diff
@@ -177,9 +177,12 @@ function getNewValidTlds($string)
...
     if (function_exists('idn_to_utf8')) {
-        return function ($domain) {
-            return idn_to_utf8($domain, 0, INTL_IDNA_VARIANT_UTS46);
-        };
+        if (defined('INTL_IDNA_VARIANT_UTS46')) {
+            return function ($domain) {
+                return idn_to_utf8($domain, 0, INTL_IDNA_VARIANT_UTS46);
+            };
+        }
+        return 'idn_to_utf8';
```
xtreamwayz:
```php
return idn_to_utf8($domain);
```

weierophinney: We don't need to do the check here, as this script is only executed by maintainers and by Travis (which has an up-to-date libicu already).

PR #203, opened 2017-10-05
Added a new type of Russian bank card - "MIR".

shlaikov:
Recently, a new type of bank card has appeared in Russia. I'd like to use it in my favorite framework.

Link to the [New Bank Card](http://mironline.ru/)


mercuriev:
Please merge it. The card type is used widely.

weierophinney:
Thanks, @shlaikov!

Issue #215, opened 2018-02-02
Bug with the Step validator


biserantonov:
Step validator is not working properly. For example if I set `baseValue = 50` and `step = 0.01` then the validation fails for the value of `50.09`, which is perfectly valid. I think the problem is in the calculation of the `fmod`:

`$fmod = $this->fmod($value - $this->baseValue, $this->step);`

In this example `$value - $this->baseValue == 0.090000000000003` and then `$fmod == 3.41E-15`. Maybe it the fmod is rounded to the precision of the step the issue will be fixed.

I'm using PHP 5.6.33

Ocramius:
Requires a test case

codisart:
Hi @Ocramius and @froschdesign,

I added some unit tests to prove the bug [pull request](https://github.com/zendframework/zend-validator/pull/218)
You can see the [travis build](https://travis-ci.org/zendframework/zend-validator/builds/343024015?utm_source=github_status&utm_medium=notification)

PR #218, opened 2018-02-18
issue-#215 - fix the step validator

codisart:
it is caused by a precision error when substracting.

### DIFF
--- a/src/Step.php
+++ b/src/Step.php
@@ -122,7 +122,9 @@ class Step extends AbstractValidator
 
         $this->setValue($value);
 
-        $fmod = $this->fmod($value - $this->baseValue, $this->step);
+        $substract = $this->sub($value, $this->baseValue);
+
+        $fmod = $this->fmod($substract, $this->step);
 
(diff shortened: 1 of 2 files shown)


src/Step.php
```diff
@@ -146,10 +148,27 @@ protected function fmod($x, $y)
...
     }
+
+    /**
+     * replaces the internal substraction operation which give wrong results on some cases
+     *
+     * @param float $x
+     * @param float $y
+     * @return float
+     */
+    protected function sub($x, $y)
```
samsonasik: should be `private`

codisart: I choose protected since fmod was in protected also. If I use private for sub, should i use private for fmod as well ?

samsonasik: I think `fmod` function should be private also, but that is BC break. Newly introduce internal function should be private imo.

codisart: I agree, I will make the modification.

codisart: done

weierophinney:
Thanks, @codisart!

PR #231, opened 2018-05-29
Correcting validation error

adamoki:
Sometimes there is a problem with validation - when generated hash is valid integer (includes only digits - it rarely happens) - php converts that string to integer - then you get integer in array of hashes ($hashes). Comparing that integer to returned string from hash_file function results in false value while hash is the same but differs in variable type.

### DIFF
--- a/src/File/Hash.php
+++ b/src/File/Hash.php
@@ -163,7 +163,7 @@ class Hash extends AbstractValidator
             }
 
             foreach ($hashes as $hash) {
-                if ($filehash == $hash) {
+                if ($filehash === $hash) {
                     return true;
                 }
             }
--- a/src/File/Hash.php
+++ b/src/File/Hash.php
@@ -163,7 +163,7 @@ class Hash extends AbstractValidator
             }
 
             foreach ($hashes as $hash) {
-                if ($filehash === $hash) {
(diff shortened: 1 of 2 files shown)


src/File/Hash.php
```diff
@@ -163,7 +163,7 @@ public function isValid($value, $file = null)
             }
 
             foreach ($hashes as $hash) {
-                if ($filehash === $hash) {
+                if ($filehash == $hash) {
```
webimpress: I think we should leave it as it is and in the other place assure that `$hash` is always string.

weierophinney:
We definitely need a unit test for this before we can merge. The unit test should fail under current master, and pass after the patch is  applied.

I tend to agree with @webimpress - we should be casting hashes to strings when calling `setHash()` and/or in the constructor.

webimpress:
@weierophinney I've pushed changes to this branch. I am checking now if input hash is string, and if not I am throwing an exception. We do have that check when we provide just a hash, without algorithm, but with algorithm we haven't had this check. Please see test I've added.

@adamoki Are you satisfied with the solution I've provided?

webimpress:
Thanks, @adamoki!

PR #237, opened 2018-08-16
PSR-7 Uploaded File compatibility

alextech:
This patch provides changes to both the `File\Upload` and `File\UploadFile` validators to provide support for validating PSR-7 `UploadedFileInterface` instances.

### DIFF
--- a/src/File/Upload.php
+++ b/src/File/Upload.php
@@ -10,6 +10,7 @@
 namespace Zend\Validator\File;
 
 use Countable;
+use Psr\Http\Message\UploadedFileInterface;
 use Zend\Validator\AbstractValidator;
 use Zend\Validator\Exception;
 
@@ -87,7 +88,11 @@ class Upload extends AbstractValidator
                     $return[$file] = $this->options['files'][$name];
                 }
 
-                if ($content['name'] === $file) {
+                if ($content instanceof UploadedFileInterface) {
+                    if($content->getClientFilename() === $file) {
+                        $return[$name] = $this->options['files'][$name];
+                    }
+                } else if ($content['name'] === $file) {
                     $return[$name] = $this->options['files'][$name];
                 }
             }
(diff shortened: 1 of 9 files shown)


samsonasik:
@alextech please rebase against develop branch to only show your commit:

```
git checkout develop
git pull https://github.com/zendframework/zend-validator.git develop
git checkout feature/psr-uploadfile
git rebase develop
```

If you found conflict, you can fix it in the file conflicted, and run: 

```
git add . && git rebase continue
```

Lastly, force push: 

```
git push --force origin feature/psr-uploadfile
```

alextech:
@samsonasik Thank you for instructions! I saved them to my notes.

src/File/Upload.php
```diff
@@ -164,9 +182,15 @@ public function isValid($value, $file = null)
...
+            $error = ($content instanceof UploadedFileInterface) ?
+                $content->getError() : $content['error'];
+            switch ($error) {
                 case 0:
-                    if (! is_uploaded_file($content['tmp_name'])) {
+                    $tmpFile = ($content instanceof UploadedFileInterface) ?
+                        $content->getStream()->getMetadata('uri') :
+                        $content['tmp_name'];
+
+                    if (! is_uploaded_file($tmpFile)) {
```
weierophinney: The above logic is incorrect.

For PSR-7, we can assume that if we have UPLOAD_ERROR_OK, then the upload performed correctly. Why? because `tmp_name` has been taken out of the equation, and we're working with streams instead.

I'll update this during merge.

test/File/UploadFileTest.php
```diff
@@ -10,6 +10,7 @@
 namespace ZendTest\Validator\File;
 
 use PHPUnit\Framework\TestCase;
+use Zend\Diactoros\UploadedFile;
```
weierophinney: I will update the patch to use PSR-7 mocks instead.

test/File/UploadTest.php
```diff
@@ -131,6 +132,117 @@ public function testBasic()
         $this->assertArrayHasKey('fileUploadErrorFileNotFound', $validator->getMessages());
     }
 
+    /**
+     * Ensures that the validator follows expected behavior
+     *
+     * @return void
+     */
+    public function testPsrBasic()
```
weierophinney: I will update this method to instead use a data provider, with mock uploads.

weierophinney:
Thanks, @alextech!

PR #238, opened 2018-09-13
Update validator-chains.md

alamaby:
The flag for `$breakChainOnFailure` should be second parameter to `attach` method, not `StringLength`.


weierophinney:
Thanks, @alamaby

Issue #240, opened 2018-10-03
Hostname validator produces notice on empty domainPart


andreasschroth:
Example input: `some-random-email@.com `

Notice + first call:
```
grapheme_strpos: Offset not contained in string---#0 /private/vendor/zendframework/zend-stdlib/src/StringWrapper/Intl.php(86): grapheme_strpos('', '-', 0)
#1 /private/vendor/zendframework/zend-validator/src/Hostname.php(2103): Zend\\Stdlib\\StringWrapper\\Intl->strpos('', '-')
```

andreasschroth:
PR proposed.

PR #241, opened 2018-10-03
Skip further checks in case of empty domain part

andreasschroth:
Fixes issue #240


andreasschroth:
One machine at Travis CI was hanging and stopped after 10 mins. The build is actually passing.

weierophinney:
Thanks, @andreasschroth!

PR #249, opened 2018-12-20
Added IDN pattern for .RS (Serbia)

nidzho:
As of Dec 10, 2018 .RS became an IDN TLD:
https://www.rnids.rs/en/domains/rs-idn-domain
https://www.rnids.rs/files/Domeni/IDN-Table-RS.pdf


weierophinney:
Thanks, @nidzho!

PR #256, opened 2019-01-23
Fix fault subdomain when not checking TLD issue #255

### LINKED ISSUE #255: Hostname validation not correct when not checking TLD
I have a use case were I check a hostname but don't want to check the tld, 

```
$validator = new \Zend\Validator\Hostname([
    'useTldCheck' => false,
    'allow' => $allow,
 ]);
$validator->isValid('unvalid_domain.be'); // true
```

This results in true, but this should be false.

SiebeVE:
Fixes #255

### DIFF
--- a/src/Hostname.php
+++ b/src/Hostname.php
@@ -2053,6 +2053,7 @@ class Hostname extends AbstractValidator
                     }
 
                     // Match TLD against known list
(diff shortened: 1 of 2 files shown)


weierophinney:
Thanks, @SiebeVE!

PR #262, opened 2019-04-09
Use pathinfo to get file extension instead of substr.

spectrox:
I think it's a bit safer to use `pathinfo()` instead of `substr()`.


webimpress:
@spectrox 

> I think it's a bit safer to use `pathinfo()` instead of `substr()`.

Can you explain it please? You _think_ based on ... ?

As I have checked it is not faster for sure:
https://3v4l.org/Be0iY

```php
$filename = 'fgdsahktyrewagfbdhzbvt74tf.gfdhsgfghsdf.gsdfhjgfdsg.gsfdghfdsh.dtasfghsdghg.tar.gz';

$t1 = microtime(true);
for ($i = 0; $i < 100000; ++$i) {
    $e1 = substr($filename, strrpos($filename, '.') + 1);
}
$t2 = microtime(true);
for ($n = 0; $n < 100000; ++$n) {
    $e2 = pathinfo($filename, \PATHINFO_EXTENSION);
}
$t3 = microtime(true);
for ($j = 0; $j < 100000; ++$j) {
    $e3 = substr(strrchr($filename, '.'), 1);
}
$t4 = microtime(true);

echo 'strrpos: ' . ($t2-$t1), PHP_EOL;
echo 'pathinfo: ' . ($t3-$t2), PHP_EOL;
echo 'strrchr: ' . ($t4-$t3), PHP_EOL;
```

Example results:
```
strrpos: 0.0073950290679932
pathinfo: 0.35552787780762
strrchr: 0.0089771747589111
```

very similar on different PHP versions.

spectrox:
Sorry, I missed your first reply.
My thinking is based on the fact that `substr`/`strrpos` functions aren't unicode-safe, while IIRC `pathinfo` is locale aware and will act normally as long as your locale will be set to the right one.

webimpress:
@spectrox can you then provide an example which is failing with `substr`/`strrpos` and is passing with `pathinfo`?

PR #263, opened 2019-04-12
Use correct mime type for jpeg files

metalinspired:
Even though most of the time jpeg files have .jpg extension the correct mime type is 'image/jpeg' as defined in https://www.w3.org/Graphics/JPEG/


webimpress:
Thanks, @metalinspired!

Issue #259, opened 2019-04-05
Proposal: Validation for breached passwords


DragonBe:
Hi,

I'm maintainer of [dragonbe/hibp](https://packagist.org/packages/dragonbe/hibp), a composer package to validate passwords against the [Have I been pwned?](https://www.haveibeenpwned.com) API service to see if a provided password was already used in a breach.

Today (April 5, 2019) [Symfony announced](https://symfony.com/blog/new-in-symfony-4-3-compromised-password-validator) they integrate the same service as a validation component in their 4.3 version and I see this as a good way to offer this functionality as part of the framework toolkit. Too bad they're not using my service though.

So I want to propose a validation rule to check provided passwords against HIBP API.

Ocramius:
@DragonBe I'd say "go ahead and implement" - this is very valuable.

As a note: please consider using a PSR-18 client as a dependency of the HIBP validator :+1:

DragonBe:
Thank you @Ocramius, I will provide a solid solution

DragonBe:
I'm contemplating on naming conventions. The service is called "Have I been pwned?", but not everyone will recognise `hibp` as a reference to this service. Symfony called it `NotPwned` but that's not in line with the naming conventions for validators.

I was thinking about the following names:
- passwordBreach
- compromisedCredentials
- commonUsedPassword

Where `passwordBreach` is my personal favourite. Any suggestions?

Ocramius:
PasswordWasNotDisclosed or such?

DragonBe:
Let's turn your proposal into a subject form "UndisclosedPassword" to match the other validators.

DragonBe:
Hey @Ocramius,

I've got a first [PoC](https://github.com/DragonBe/zend-validator/commit/ddc6fa97c23b078de38d648557e91faebd468bef) as a test provided. In this use case I'm using my own [hibp package](https://packagist.org/packages/dragonbe/hibp) to prevent maintaining 2 versions. I need to update my package first to ensure it implements PSR-7 and PSR-18 so users can make use of their own clients instead of the provided GuzzleHttp that's part of my package. Is this an approach you find OK or do you prefer to have all that logic build as a separate zend component?

How far do you want this component to deliver validation? The basic functionality is to use the provided API by Have I been pwned, but there's also an offline validation possible using downloaded password hash files. This functionality is for users who want to prevent sending password hash snippets over the wire.

DragonBe:
[...]

PR #264, opened 2019-04-13
#259 Validation for breached passwords

DragonBe:
This is the UndisclosedPassword validator discussed in #259 which uses PSR-18 http client interface for communicating with [Have I Been Pwned?](https://www.haveibeenpwned.com) web service for validating passwords in a secure, safe way using K-Anonymity method.

**IMPORTANT:** The thing that might be a problem for this pull request is that PSR-18 composer package requires PHP 7.0 or higher.

All functionality is documented and provided with unit tests ensuring 100% code coverage.
image

Please review this pull request and let me know if I need to change or improve the functionality.


composer.json
```diff
@@ -30,7 +30,9 @@
         "zendframework/zend-math": "^2.6",
         "zendframework/zend-servicemanager": "^2.7.5 || ^3.0.3",
         "zendframework/zend-session": "^2.8",
-        "zendframework/zend-uri": "^2.5"
+        "zendframework/zend-uri": "^2.5",
+        "psr/http-client": "^1.0",
```
Ocramius: These should be near the other `psr/*` libs above

src/UndisclosedPassword.php
```diff
@@ -0,0 +1,177 @@
...
+    const HIBP_COUNT_BASE = 0;
+    const SHA1_LENGTH = 40;
+
+    const PASSWORD_BREACHED = 'passwordBreached';
+    const WRONG_INPUT = 'wrongInput';
+    const CONNECTION_FAILURE = 'connFail';
+    const UNKNOWN_ERROR = 'unknownError';
+
+    protected $messageTemplates = [
+        self::PASSWORD_BREACHED => 'The provided password was used by others',
```
Ocramius: Reading this in the UI can be confusing for a user: who is "others" to them?

src/UndisclosedPassword.php
```diff
@@ -0,0 +1,177 @@
...
+        self::WRONG_INPUT => 'The provided password failed to be correctly hashed, please verify your input',
```
Ocramius: Given usages of `WRONG_INPUT`, I'd say that:

 * the constant should be renamed to `NOT_A_STRING` (or something like that)
 * the message should reflect the same

src/UndisclosedPassword.php
```diff
@@ -0,0 +1,177 @@
...
+        self::CONNECTION_FAILURE => 'Unable to reach HIBP service, please try again later',
```
Ocramius: HIBP is a text that doesn't make sense for the typical end-user: we'll need a better error message here too

src/UndisclosedPassword.php
```diff
@@ -0,0 +1,177 @@
...
+        self::UNKNOWN_ERROR => 'Something happened beyond our control, error reporting should give more details',
```
Ocramius: Also here: this error message is developer-friendly, but not end-user friendly

src/UndisclosedPassword.php
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * @var ResponseFactoryInterface
+     */
+    private $httpResponse;
+
+    /**
+     * @var int
+     */
+    private $count = 0;
```
Ocramius: What count is this representing? Looking just at the property, this can be confusing

src/UndisclosedPassword.php
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * @var ClientInterface
+     */
+    private $httpClient;
+
+    /**
+     * @var RequestFactoryInterface
+     */
+    private $httpRequest;
```
Ocramius: s/httpRequest/makeHttpRequest maybe? (since this isn't a real request, but the factory for it)

src/UndisclosedPassword.php
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * @var ResponseFactoryInterface
+     */
+    private $httpResponse;
```
Ocramius: s/httpResponse/makeHttpResponse maybe? (since this isn't a real response, but the factory for it)

src/UndisclosedPassword.php
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * @var int
+     */
+    private $count = 0;
+
+    /**
```
Ocramius: Docblock can be omitted, as it doesn't contain further/useful information

src/UndisclosedPassword.php
```diff
@@ -0,0 +1,177 @@
...
+        if (! is_string($value)) {
+            $this->error(self::WRONG_INPUT);
+            return false;
+        }
+        try {
+            $isPwnd = $this->isPwnedPassword($value);
+        } catch (Exception\RuntimeException $runtimeException) {
+            $this->error(self::CONNECTION_FAILURE);
+            return false;
+        } catch (\Exception $exception) {
```
Ocramius: I don't think we should catch `\Exception` here: something went really wrong here, maybe in service wiring, so we should let the exception bubble up instead, making developers aware of the problem

src/UndisclosedPassword.php
```diff
@@ -0,0 +1,177 @@
...
+     * @return string
+     * @throws Exception\RuntimeException
+     */
+    private function retrieveHashList($passwordRange)
+    {
+        $request = $this->httpRequest->createRequest('GET', '/range/' . $passwordRange);
+
+        try {
+            $response = $this->httpClient->sendRequest($request);
+        } catch (ClientExceptionInterface $connectException) {
```
Ocramius: You are catching/re-throwing and then catching in the caller again here: if we can recognise precise `ClientExceptionInterface` types, let's react to them in the caller, and otherwise let exceptions bubble up, since otherwise misconfiguration scenarios become extremely hard to inspect.

src/UndisclosedPassword.php
```diff
@@ -0,0 +1,177 @@
...
+            return false;
+        }
+        if ($isPwnd) {
+            $this->error(self::PASSWORD_BREACHED);
+            return false;
+        }
+        return true;
+    }
+
+    private function isPwnedPassword($password)
```
Ocramius: Noticed while looking at the lack of type declarations: new code should target `develop` (target branch needs changing in the PR)

src/UndisclosedPassword.php
```diff
@@ -0,0 +1,177 @@
...
+        try {
+            $isPwnd = $this->isPwnedPassword($value);
+        } catch (Exception\RuntimeException $runtimeException) {
+            $this->error(self::CONNECTION_FAILURE);
+            return false;
+        } catch (\Exception $exception) {
+            $this->error(self::UNKNOWN_ERROR);
+            return false;
+        }
+        if ($isPwnd) {
```
Ocramius: This variable is redundant - should be removed

src/UndisclosedPassword.php
```diff
@@ -0,0 +1,177 @@
...
+     * Creates a hash range that will be send to HIBP API
+     * applying K-Anonymity
+     *
+     * @param string $passwordHash
+     * @return string
+     * @see https://www.troyhunt.com/enhancing-pwned-passwords-privacy-by-exclusively-supporting-anonymity/
+     */
+    private function getRangeHash($passwordHash)
+    {
+        $range = substr($passwordHash, self::HIBP_RANGE_BASE, self::HIBP_RANGE_LENGTH);
```
Ocramius: Variable redundant

src/UndisclosedPassword.php
```diff
@@ -0,0 +1,177 @@
...
+use Psr\Http\Message\RequestFactoryInterface;
+use Psr\Http\Message\ResponseFactoryInterface;
+
+final class UndisclosedPassword extends AbstractValidator
+{
+    const HIBP_API_URI = 'https://api.pwnedpasswords.com';
+    const HIBP_API_TIMEOUT = 300;
+    const HIBP_CLIENT_UA = 'zend-validator';
+    const HIBP_CLIENT_ACCEPT = 'application/vnd.haveibeenpwned.v2+json';
+    const HIBP_RANGE_LENGTH = 5;
```
Ocramius: Looking at usages of the constants below, this makes kinda sense, but the naming of the constants themselves do not always make sense. If we can expand the names, that would be best.

src/UndisclosedPassword.php
```diff
@@ -0,0 +1,177 @@
...
+namespace Zend\Validator;
+
+use Psr\Http\Client\ClientExceptionInterface;
+use Psr\Http\Client\ClientInterface;
+use Psr\Http\Message\RequestFactoryInterface;
+use Psr\Http\Message\ResponseFactoryInterface;
+
+final class UndisclosedPassword extends AbstractValidator
+{
+    const HIBP_API_URI = 'https://api.pwnedpasswords.com';
```
Ocramius: @weierophinney all of these constants should be `private` in order to prevent needless BC breaks later on: can we please bump to `^7.1` before expanding the API surface beyond what should be `public`?

test/UndisclosedPasswordTest.php
```diff
@@ -0,0 +1,233 @@
...
+        $this->assertFalse($this->validator->isValid(['foo']));
+    }
+
+    /**
+     * Test that a given password was not found in the HIBP
+     * API service.
+     *
+     * @param string $password
+     *
+     * @covers \Zend\Validator\UndisclosedPassword::__construct
```
Ocramius: Consider using a single `@covers \Zend\Validator\UndisclosedPassword` at class level: this sort of fine-grained coverage is more added maintenance work when someone splits a `private` detail into more `private` methods

Ocramius (changes requested):

composer.json
DragonBe: @Ocramius I can do that… but have you seen my request in regards PSR-18 not being compatible with PHP 5.6? ➡️ https://github.com/zendframework/zend-validator/issues/259#issuecomment-482839690

Ocramius: Yeah, I think we need to bump the dependency here: it is long overdue and should just be done.

Also, I'm using this occasion to re-express to @zendframework/community-review-team the fact that bumping PHP version can be freely done in minor releases.

src/UndisclosedPassword.php
DragonBe: Count is used for providing a "count" of amount of times a password was found in Have I been pwned. Each hash that returns contains a number. See https://haveibeenpwned.com/API/v2#SearchingPwnedPasswordsByRange for details.

Ocramius: > amount of times a password was found

`pwnedPasswordsFoundCount`?

src/UndisclosedPassword.php
DragonBe: Even though not providing useful information beyond the default value being provided, IDE's tend to enjoy this and PHPDocumentor still uses it to generate good API documentation from it.

I would like to keep it in if you don't mind.

Ocramius: API documentation and IDE document this from the type declaration. If the docblock does not provide information about a symbol that isn't already part of the reflection of said symbol, it is noise. I don't have a hard stance on this, but in `doctrine/coding-standard`, this sort of docblock is explicitly forbidden, for example.

composer.json
Ocramius: I raised https://github.com/zendframework/maintainers/issues/59 about this

src/UndisclosedPassword.php
DragonBe: Or I could implement `Countable` so you could count on the object directly.

Ocramius: I don't think that would clarify it: it would make it even more confusing instead. Why would a `Countable` implemented on a `ValidatorInterface` implementation returning a "count of pwned found passwords"? This is just about communicating what a specific `private` state is meant for, so we shouldn't expand the `public` API of the object for an implementation detail that is already ambiguous in `private` scope.

src/UndisclosedPassword.php
DragonBe: Moved base to `develop` but even there the composer refers to php 5.6 so until your proposal is agreed upon, let's leave out the type declarations for now. It's a jiffy to put them in, but I'm already struggling with a failed pass on PHP 5.6 because of PSR18 and I'd like to tackle one issue at the time. Let's solve zendframework/maintainers#59 first before adding type declarations.

src/UndisclosedPassword.php
DragonBe: Good point, in a "validation" context it has no real purpose. I'll remove it.

doc/book/validators/undisclosed-password.md
```diff
@@ -0,0 +1,38 @@
...
+> ```
+
+## Basic usage
+
+To validate if a password was disclosed in a known data breach, you need to provide a HTTP Client that implements `psr\http\ClientInterface`, a `Psr\Http\Message\RequestFactoryInterface` and a `Psr\Http\Message\ResponseFactoryInterface` to the constructor and validate the password you want to check.
+
+If the password was found via the service, `isValid` will return FALSE. If the password was not found, `isValid` will return TRUE.
+
+```php
+$validator = new Zend\Validator\UndisclosedPassword(
```
froschdesign: The example must be completely reproducible for the end user. At the moment, the example shows the same thing as the API, which is not very helpful. We need a full code example including a HTTP client and the factories.

doc/book/validators/undisclosed-password.md
```diff
@@ -0,0 +1,38 @@
...
+> ```bash
+> $ composer require psr/http-client
+> $ composer require psr/http-factory
+> ```
+
+## Basic usage
+
+To validate if a password was disclosed in a known data breach, you need to provide a HTTP Client that implements `psr\http\ClientInterface`, a `Psr\Http\Message\RequestFactoryInterface` and a `Psr\Http\Message\ResponseFactoryInterface` to the constructor and validate the password you want to check.
+
+If the password was found via the service, `isValid` will return FALSE. If the password was not found, `isValid` will return TRUE.
```
froschdesign: Also set "false" and "true" in backticks and use lowercase letters. Thanks!

froschdesign (changes requested):
First: 👍 for adding documentation.
Second: I added some small comments. The important part is the code example because we need a full working example.

doc/book/validators/undisclosed-password.md
DragonBe: I will update this piece of example code after I figure out which PSR-18 http client I could use in this example. 

I was thinking about using the Guzzle6 adapter and factories, because zend-diactoros doesn't implement PSR-18 (yet), but I can go any direction.

DragonBe: @froschdesign In commit 1383af3 I have provided a working example of `Zend\Validator\UndisclosedPassword` using PSR7, PSR17 and PSR18 HTTP Client and Messages.

Please have a look at it if this is in line with your expectations. If not, please indicate how I can improve it.

weierophinney:
Thanks, @DragonBe! With the imminent transition to the Laminas Project, I am expecting we will likely revisit the ability to bump a PHP version during a major series, and, as such, decided to do so here, as it will ensure we do not have any BC issues with your new class in the future. I added visibility to all constants, and added typehints to parameters and return values of all internal methods. Will release shortly with version 2.13.0.

PR #266, opened 2019-04-29
fix: Make File\Extension usable for non-existing files

edannenberg:
Currently the validator fails if the given filename is not readable. This prevents valid use cases like checking if a user configurable filename has an allowed file extension before creating said file.

As the validator only operates on the given string there is no need for the current readable check.

Also fixes File\ExcludeExtension which suffers from the same issue.

In theory it's a BC breaking change, personally I would consider any code that (ab)uses this validator as a file_is_readable validator to be smelly/buggy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file, in reverse 
 
 ### Added
 
-- Nothing.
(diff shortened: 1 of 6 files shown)


weierophinney:
This is definitely a change in behavior; we have tests that verify that the validator returns `false` when the file does not exist (which you can see in the [test results](https://travis-ci.org/zendframework/zend-validator/jobs/525934909#L662)).

That said, looking through the code, you are absolutely correct - the validators are not testing any of the file _contents_, just the file _name_.

For BC purposes, this needs to be opt-in behavior, with the current behavior being the default. We can revisit that when we consider a new major version.

I'll push a change shortly to enable this.

weierophinney:
I've pushed a change to make this optional behavior, and added documentation on how to enable the option.

weierophinney (approved):
With the change to make the new behavior optional, this can go into the next minor.

PR #272, opened 2019-08-27
Fix: replace curly offset access brace with square brackets

webimpress:
As of PHP 7.4:
the array and string offset access syntax using curly braces is deprecated.

### DIFF
--- a/src/Isbn/Isbn10.php
+++ b/src/Isbn/Isbn10.php
@@ -32,7 +32,7 @@ class Isbn10
         $sum = 0;
 
         for ($i = 0; $i < 9; $i++) {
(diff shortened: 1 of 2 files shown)


froschdesign:
Reference: https://wiki.php.net/rfc/deprecate_curly_braces_array_access

samsonasik:
`7.4snapshot` need to be added to `.travis.yml`

froschdesign:
@samsonasik 
> `7.4snapshot` need to be added to `.travis.yml`

We need a strategy for this, otherwise a new problem will arise at the end of November: we need to change the configuration for Travis again - manually.

PR #277, opened 2019-10-28
Fixes File\Hash validator to work with digit-only hashes and match algo

### LINKED ISSUE #276: Hash and Crc32 validator issue if hash has only digits
I've found an issue in Hash::isValid() as well as in Crc32::isValid() methods. If tested file's hash has only digits, then in
```php
$hashes = array_unique(array_keys($this->getHash()));
```
$hashes array has value which is an integer. And because hash_file returns string then 
```php
if ($filehash === $hash) {
```
returns false.

This is example file, which has digits only crc32 hash
[0284a0f98c9787df3302fd7c9e6fd47d.pdf](https://github.com/zendframework/zend-validator/files/3763361/0284a0f98c9787df3302fd7c9e6fd47d.pdf)

I think, that solution is simple - casting to string before comparison:
```php
        foreach ($hashes as $hash) {
            if ((string)$filehash === $hash) {
                return true;
            }
        }
```

webimpress:
Validation was failing when file hash was digit-only due to issue with the implementation and silent type conversion when using `array_keys` on array with digit-only string keys.

Also fixed other issue, where hash was not checked against the algorithm.

Fixes #276

Provide a narrative description of what you are trying to accomplish:

- [x] Are you fixing a bug?
  - [x] Detail how the bug is invoked currently.
  - [x] Detail the original, incorrect behavior.
  - [x] Detail the new, expected behavior.
  - [x] Base your feature on the `master` branch, and submit against that branch.
  - [x] Add a regression test that demonstrates the bug, and proves the fix.
  - [ ] Add a `CHANGELOG.md` entry for the fix.


froschdesign:
@webimpress 
The failure is not related to this PR:

```
1) ZendTest\Validator\EmailAddressTest::testNoMxRecordARecordFallback
Email host contains MX records
Failed asserting that true is false.
/home/travis/build/zendframework/zend-validator/test/EmailAddressTest.php:404
```

https://travis-ci.org/zendframework/zend-validator/jobs/604109135#L428

webimpress:
@froschdesign 
> The failure is not related to this PR

Yes, I've seen it before, it looks like `example.com` domain has now defined some MX records, but before it hasn't. Because of that some tests are failing. We need to sort it out somehow - maybe our custom subdomain without mx records?

